PR #1164, opened 2012-05-26
Added the ability to do codemirror true on a text field and have a codem...

rovermicrover:
Sorry about the one before forgot to add a topic branch and rebase everything. Tested again with "bundle exec rake spec" and worked fine.

---

These commits should allow for something like this.

``` ruby
field :css, :text do
    codemirror true
end
```

requires
https://github.com/fixlr/codemirror-rails

The commits also add the following options for setting up ckeditor.

``` ruby
#Use this if you want to point to a cloud instances of CKeditor
register_instance_option(:ckeditor_location) do
  '/assets/ckeditor/ckeditor.js'
end

#Use this if you want to point to a cloud instances of the base CKeditor
register_instance_option(:ckeditor_base_location) do
  '/assets/ckeditor/'
end
```

Then for Codemirror the options are

``` ruby
# Codemirror is disabled by default and CKEditor takes presedence
register_instance_option(:codemirror) do
  false
end

#Pass the theme and mode for Codemirror
register_instance_option(:codemirror_config) do
  {
    :mode => 'css',
    :theme => 'night'
  }
end

register_instance_option(:codemirror_assets) do
  {
    :mode => '/assets/codemirror/modes/css.js',  
    :theme => '/assets/codemirror/themes/night.css'
  }
end

#Use this if you want to point to a cloud instances of CodeMirror
register_instance_option(:codemirror_js_location) do
  '/assets/codemirror.js'
end

#Use this if you want to point to a cloud instances of CodeMirror
register_instance_option(:codemirror_css_location) do
  '/assets/codemirror.css'
end
```

Some Notes:
I had to remove @import "bootstrap/code"; from imports.css.scss.erb because it was conflicting with the basic code mirror styling.

I have tested 

``` ruby
codemirror true
ckeditor true
```

Both still work with one or more instances.

I have though had a tough time testing the configs I added. I guess I must not be setting them correctly or something because I can't get them to be anything but the default.

Inspired by:
The following issue posted by https://github.com/kennyadsl
https://github.com/sferik/rails_admin/issues/871

Implementation via binding on a project I worked a little bit on with
https://github.com/nthj

I hope this helps!

### DIFF
--- a/app/assets/javascripts/rails_admin/ra.widgets.coffee
+++ b/app/assets/javascripts/rails_admin/ra.widgets.coffee
@@ -125,11 +125,23 @@ $(document).live 'rails_admin.dom_ready', ->
     # ckeditor
 
     $('form [data-richtext=ckeditor]').not('.ckeditored').each ->
-      window.CKEDITOR_BASEPATH = '/assets/ckeditor/'
       options = $(this).data('options')
+      window.CKEDITOR_BASEPATH = options['base_location']
       if not window.CKEDITOR
         $(window.document).append('<script src="' + options['jspath'] + '"><\/script>')
       if instance = window.CKEDITOR.instances[this.id]
         instance.destroy(true)
       window.CKEDITOR.replace(this, options['options'])
       $(this).addClass('ckeditored')
+
+    #codemirror
+
+    $('form [data-richtext=codemirror]').not('.codemirrored').each ->
+      options = $(this).data('options')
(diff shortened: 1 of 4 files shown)


rikkipitt:
Hey guys, are multiple instances of CodeMirror doable on the same page? Example: a css and js field. It seems like only the second code mirrored field works, are the settings overridden by the second field?

Issue #1203, opened 2012-06-13
Object ID attribute is changed when updating object


rzmz:
Hi, I found something that might be an issue, not sure though. Somebody more experienced should look into this perhaps.
Scenario: using Devise (with CanCan) also for application authentication. When trying to edit my User object in Rails Admin, always the current user logged in gets updated instead (the :id attribute was being changed in lib/rails_admin/config/actions/edit.rb) 

I monkeypatched that issue in a fork but I am not sure, why or where the :id attribute is added for changing.
https://github.com/rzmz/rails_admin/commit/247633ba71932bb64a33f0fa4ebc85167071976a

PR #1243, opened 2012-07-11
Don't let authorization_adapter overwrite id when creating or updating an object

christophermanning:
This is related to Issue #1203

If a permission like `can :manage, Player, :id => 2` was set (when you want to only allow a user to manage certain Players) it would overwrite the id of the object and could cause an update of the wrong object or creating an object with a duplicate primary key.

### DIFF
--- a/spec/integration/authorization/cancan_spec.rb
+++ b/spec/integration/authorization/cancan_spec.rb
@@ -9,7 +9,7 @@ class Ability
       can :manage, Player if user.roles.include? :manage_player
       can :read, Player, :retired => false if user.roles.include? :read_player
       can :create, Player, :suspended => true if user.roles.include? :create_player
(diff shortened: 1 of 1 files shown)


cec:
Hi @bbenezech,

I think that this issue happens when using a role-based authorisation model with hierarchical models: higher roles inherit abilities from lower roles.

For example, imagine two roles: guest, admin.
Where guest users can :edit their own user only (there is a `can` rule matching on their id).
Admin users can manage all other users.

In this case, if admin *a* tries to edit a user *u1*, authorization_adapter will end up trying to set the id of *u1* to the id of *a*.

To me, a scenario where guests can "edit their profile" and admins can do anything is common sense.

How else should I handle this? Should I break the permission inheritance logic for this particular case?

If I have not expressed my situation clearly, please let me know and I will attach my ability file too.

Thanks a lot!

PR #1363, opened 2012-10-18
Add filter on an HABTM associated model's field

Val:
Template to test how it works : https://gist.github.com/3810559.

Hope it helps to provide this functionality.


mshibuya:
As is stated in https://github.com/sferik/rails_admin/issues/1342#issuecomment-9766623, we need some other way around.

Issue #1434, opened 2012-12-10
has_many :through associations filter problem


barsukov:
Hi ,I have a problem with adding has many through association to filter . When I try to add this code :
```
list do
    field :model ,:has_many_association do
          searchable :name
     end
end'
```
and then I add choose filter on the view I give the exception : 

Unknown column 'model.name' in 'where clause': SELECT  `another_model`.* FROM `another_model`  WHERE (((model.name_en LIKE '%test%')))

The same problem I found on the StackOVerflow but it has no answer :( 
http://stackoverflow.com/questions/13529634/rails-admin-searchable-association.
Could you help me to resolve it?

vedant1811:
Is the issue still there? there seems to be an easy fix as per the blog of @barsukov (issue opener): http://blog.endpoint.com/2013/07/hasmany-filter-in-railsadmin.html

barsukov:
Yes if it imlements you can close but I do not remember this issue it has been 2 years ago :smile:

rafaelgonzalez:
The issue is still there. Just had a try with `master`.

I have implemented the solution on that blog post as well, more specifically the one in the comments. The solution in the post does not work for `master`.

vincentwoo:
@sferik any interest in addressing this issue? seems to be a pretty useful feature to have

mshibuya:
@vincentwoo Of course, as long as it does not break existing functionality.
Please refer to the note in https://github.com/sferik/rails_admin/pull/2349#issuecomment-118215266.

vincentwoo:
@mshibuya It seems simple enough to only search `has_many` associations where the programmer has explicitly enabled `searchable` on, with a disclaimer about the performance tradeoffs. In the case that I'm interested in, most records will have 1-4 objects in the association, which is fine for performance.

mshibuya:
Sounds good, could you submit a PR, along with tests?

vincentwoo:
@mshibuya Unfortunately I don't think I have the time/expertise to tackle this right now, but if anyone wants to I'm happy to kick the tires!

fmh:
@rafaelgonzalez, it still does not work for me, even with your solution,
RA , master or 0.7
````ruby
class Order < AR
 has_many :order_clients
 has_many :clients, through: :order_clients
 rails_admin do 
   field :clients do
          searchable :last_name
      end
 end
end

[...]

PR #1437, opened 2012-12-15
Issue #1434 filter functionality for has many association.

barsukov:
This commit has few changes : 
1) I implement filter functionality for has_many association with issue #1434
2) I refactor one spec with rspec best practices and factory girl too.


rafaelgonzalez:
@bbenezech 

It makes sense to me to have a HM or HABTM filter.

I have a `Project` model with many `Task` and want to be able to filter out projects with tasks with a specific name, or a name that has certain content. (in truth it is really a HABTM relationship in my case, so I would like to filter all the projects that have the same task assigned to them but I use `has_many :through` with an intermediate model)

Please see #1434 were a partial solution has been discussed.

barsukov:
I think this is not actual anymore?

rafaelgonzalez:
Well the solution I use in #1434 only partially solves my problem (see my comment above), so to me this is still actual.

mshibuya:
The same as #1363.

PR #1610, opened 2013-04-19
polymorphic associations are scopeable

zambot:
This patch enables the scope for a polymorphic association to be set (in the same way as it already works in the non polymorphic associations).
It enables caching if a scope is set.


bbenezech:
staled, to many conflicts.

@coveralls Feel free to reopen it if you want to rebase it. Thanks!

PR #1749, opened 2013-09-19
Support Single Table Inheritance in paper trail history

aceofspades:
History is not shown for subclasses when using Single Table Inheritance with paper_trail, since paper_trail stores the name of the base class in the versions table.


bbenezech:
I am closing this as stalled. Feel free to reopen if the need arises.

PR #1753, opened 2013-09-20
Allow proc for static_navigation url parameter

aceofspades:
Handle a Proc vs. String as a navigation_static_link url


halilim:
:+1:

mshibuya:
Merging to #2345.

PR #1781, opened 2013-10-04
Allow multiple config blocks for a single resource in RailsAdmin

sepastian:
When using RailsAdmin within another gem, it should be possible to spread out and merge configuration blocks for a single resource across initializers and models, in an attempt to produce DRY code.

Currently, RailsAdmin allows multiple configuration blocks for a single resource, however, blocks evaluating later will overwrite earlier blocks.

This pull requests contains two small changes to `RailsAdmin::Config::LazyModel` and `RailsAdmin::Config`, to implement multiple configuration blocks for a single resource.

For example, in the current version of RailsAdmin, these two blocks would overwrite each other - depending on the exact order of loading (see comment in `lib/rails_admin/config/lazy_model.rb`) :

    # app/models/some_model.rb
    class SomeModel
      rails_admin do
        :
      end
    end

    # config/initializers/rails_admin.rb
    config.model SomeModel do
      :
    end

After applying this pull request, the settings defined in `config/initializers/rails_admin.rb` will evaluate before the ones found in `SomeModel`. The settings from `config/initializers/rails_admin.rb` will be maintained, although `SomeModel` may chose to overwrite some settings.

### DIFF
--- a/lib/rails_admin/config.rb
+++ b/lib/rails_admin/config.rb
@@ -237,11 +237,9 @@ module RailsAdmin
           end
         end
 
-        if block
-          @registry[key] = RailsAdmin::Config::LazyModel.new(entity, &block)
-        else
-          @registry[key] ||= RailsAdmin::Config::LazyModel.new(entity)
-        end
+        @registry[key] ||= RailsAdmin::Config::LazyModel.new(entity)
+        @registry[key].add_deferred_block(&block) if block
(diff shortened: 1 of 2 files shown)


bbenezech:
Closing this as stalled. Please reopen and rebase if you still want to add it. Some tests would definitely be needed as well for such a feature.

PR #1944, opened 2014-03-30
fixing bug undefined method `abstract_model' for

mido18:
fixing bug in polymorphic belongs_to has_one relationship


mshibuya:
What kind of trouble do you have?
If you still need this fix, please reopen with tests attached.

Issue #1665, opened 2013-06-06
Disable navigation_static_label


ricardokrieg:
It would be good to have a way to disable the **navigation_static_label**

With the current code, there's no way to disable it:

```
label = RailsAdmin::Config.navigation_static_label || t('admin.misc.navigation_static_label')
```

Maybe creating a new config param:

```
config.disable_navigation_static_label = true
```

And changing the code at **static_navigation** to this:

```
unless RailsAdmin::Config.disable_navigation_static_label
  label = RailsAdmin::Config.navigation_static_label || t('admin.misc.navigation_static_label')
  li_stack = %{<li class='nav-header'>#{label}</li>#{li_stack}}.html_safe if li_stack.present?
end
```

PR #1964, opened 2014-04-15
Disable navigation_static_label

rafaellima:
Based on issue [#1665](https://github.com/sferik/rails_admin/issues/1665).
It introduces an option to RailsAdmin::Config ```disable_navigation_static_label``` that doesn't show navigation_static_label when set as true.


mshibuya:
#2345 will handle this.

PR #1965, opened 2014-04-15
Add instance option 'confirmation_required' that will require that an option shows a confirmation prompt

jordanmaguire:
Hi team,

Is this feature something that you would consider incorporating into RailsAdmin? A confirmation prompt seems like something that would be commonly required - I certainly need it in the app I'm building, currently.

If you are ok with the idea of this feature, I plan to add in some specs and incorporate I18n so that you can customise the confirmation prompt per action/model.

### DIFF
--- a/lib/rails_admin/config/actions/base.rb
+++ b/lib/rails_admin/config/actions/base.rb
@@ -18,6 +18,11 @@ module RailsAdmin
           []
         end
 
+        # Should the action show a confirmation (using jquery_ujs) before performing this action
+        register_instance_option :confirmation_required do
+          false
+        end
+
         # http://getbootstrap.com/2.3.2/base-css.html#icons
         register_instance_option :link_icon do
           'icon-question-sign'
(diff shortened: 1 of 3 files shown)


lib/rails_admin/config/actions/base.rb
```diff
@@ -18,6 +18,11 @@ class Base
           []
         end
 
+        # Should the action show a confirmation (using jquery_ujs) before performing this action
+        register_instance_option :confirmation_required do
+          false
+        end
```
mshibuya: Changing the name to `:confirmation` and letting users set customized message will be more flexible, IMHO.

bbenezech:
I think it is better to add it with rails-ujs. You can override `rails_admin/ui.js` and hook into the dom with `#{action.key}_#{parent}_link`

jordanmaguire:
That makes sense.

I've added the following coffeescript to do this:

```coffeescript
$(document).on 'rails_admin.dom_ready', ->
  if ($archiveButtons = $(".archive_member_link a")).length
    $archiveButtons.click (event) ->
      # Returning false will prevent the link from being followed
      return confirm("Are you sure you want to archive this item?")

PR #1983, opened 2014-05-08
Model label should be pluralized with locale

Kabie:
Right now it uses `locale: :en`. So it might be incorrectly pluralized.


mshibuya:
Could you please write some test?

bbenezech:
Good catch!
@mshibuya Current behavior test should be more than enough.
>> http://apidock.com/rails/ActiveSupport/Inflector/pluralize

PR #2004, opened 2014-06-09
const_defined does not work with namespaced models

RafaelAlba:
If history is enabled and a namespaced model gets deleted, this throws a "wrong constant name" error message.

I added the .try just to be extra safe.


mshibuya:
Test needed!

bbenezech:
Pb has been solved in Ruby 2 years ago :disappointed_relieved: 
Very sorry for this.

PR #2007, opened 2014-06-11
dependent_field for belongs_to

fantgeass:
Dependent field for belongs_to association.

Example:

We have user with country/city.
User chooses country "Russia" and city field now contains all russian cities.
When user changes country  to "France", city field will lost value and select list now contain all "France" cities.
~~~
class User < ActiveRecord::Base
	belongs_to :country
	belongs_to :city

	rails_admin do
		edit do
		  field :city do
		    dependent_field :country
		    associated_collection_scope do
		      country_id = bindings[:controller].params[:dependent_field_value]
		      Proc.new { |scope|
		        scope = scope.where(country_id: country_id) if country_id
		      }
		    end
		  end
		end
  end
end
~~~

### DIFF
--- a/app/assets/javascripts/rails_admin/ra.filtering-select.js
+++ b/app/assets/javascripts/rails_admin/ra.filtering-select.js
@@ -114,7 +114,16 @@
 
       filtering_select.append(input).append(button).insertAfter(select);
 
+      if (this.options.dependent_field_selector) {
+        self.options.dependent_field_value = $(this.options.dependent_field_selector).val();
+        $(this.options.dependent_field_selector).on('change', function () {
+          self.options.dependent_field_value = $(this).val();
 
+          input.val('');
+          select.html($('<option value="" selected="selected"></option>'));
+          select.trigger("change");
+        })
+      }
     },
 
     _getResultSet: function(request, data, xhr) {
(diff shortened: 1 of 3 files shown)


mshibuya:
Please add some tests.

SuperMasterBlasterLaser:
This feature seems to be very useful. I solved it by creating custom javascript code that will filter autocomplete values depending on value of other select. Also this required to me to create controller.

SuperMasterBlasterLaser:
@fantgeass Does this dependent_field work with `has_and_belongs_to_many` association where filtering multiselect will be?

chrisbloom7:
I came up with an [alternative solution that wasn't too involved](http://www.christopherbloom.com/2014/09/04/dynamic-enum-fields-in-nested-association-forms-in-rails-admin/). Having it baked in to RA as an option would be very useful, but my approach isn't that much more work and allows you to programmatically determine what subitems should be displayed beyond simply filtering.

SuperMasterBlasterLaser:
@chrisbloom7 This solution works well, but it is for specific conditions. You can create your own controller that interracts with `ui,js` javascript code.

However, we need universal filtering, which will be configured via `field` in `rails_admin.rb`. It must be able to work with any `has_many` and `has_and_belongs_to_many` associations. Also it must not require additional controller from user.

chrisbloom7:
@SuperMasterBlasterLaser agreed. Was just listing an alternative for those looking to do so before it's part of RA core functionality.

bbenezech:
IMO this should not be part of RA core.
This is way too specific, sorry.

PR #2010, opened 2014-06-16
fix parsing datetime for locales

lksv:
fix datetime to normalize correctly even when one month name (or weekday
name) is substring of another.

It is at least in Czech:
* June (červen) is a substring of Juny (červenec)

I did not create a tests. It should be done like:

    it 'is able to read %B %d, %Y %H:%M for cs locate' do
      # in Czech June (červen) is a substring of Juny (červenec)
      @time = ::Time.new(2014,7,21,23,45).getutc
      I18n.with_locale(:cs) do
        puts I18n.l @time
        @object = FactoryGirl.create(:field_test)
        @object.datetime_field = @field.parse_input(datetime_field: @time.strftime('%B %d, %Y %H:%M'))
        expect(@object.datetime_field.strftime('%Y-%m-%d %H:%M')).to eq(@time.strftime('%Y-%m-%d %H:%M'))
      end 
    end 

But it is not good idea to include Czech locale to the project. What is the best way to test it?

### DIFF
--- a/lib/rails_admin/config/fields/types/datetime.rb
+++ b/lib/rails_admin/config/fields/types/datetime.rb
@@ -24,16 +24,16 @@ module RailsAdmin
                   case match
                   when '%A'
                     english = I18n.t('date.day_names', locale: :en)
-                    day_names.each_with_index { |d, i| date_string = date_string.gsub(/#{d}/, english[i]) }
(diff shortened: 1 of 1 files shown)


mshibuya:
Spec needed!

mshibuya:
Stale.

PR #2014, opened 2014-06-24
Correctly display enums consisting of arrays in the show page

sildur:
The show page displays correctly enums when they are hashes or arrays of arrays, but it incorrectly displays enums consisting simply of arrays, such as:
```ruby
['value', 'other_value']
```

This is the form used in Rails enums. This patch correctly displays such enums.


mshibuya:
Fixed by 309a2d1e.

PR #2027, opened 2014-07-19
`pluralize` conflicts with ActionView helper

salamagd:
In Rails 4.1.4 / Ruby 2.1.x


mshibuya:
> including ActionView::Helpers::TextHelper at the global level

That's the cause of problem, I strongly recommend not to do so.
If you want helpers available in your models, [active_decorator](https://github.com/amatsuda/active_decorator) might help.

chungphuoc:
This seem not fixed yet, I use lastest versions 0.7.0 and still meet this problem

PR #2039, opened 2014-08-07
fix datetime picker

nazarok:
If I edit a field with type: DateTime everything is stored properly.
If I don't edit a time input - time is set to '00:00', because there is no change event.


mshibuya:
ra.datetimepicker was replaced with Bootstrap datetimepicker by #2391.

PR #2041, opened 2014-08-10
add option for disabling a list page's title attribute for a given colum...

superplussed:
The issue is that when adding custom formatting to a field into a list page, the title attribute can turn pretty unsightly as shown in this screenshot:

title_attribute

So I added a config option on a property called 'hide_title_attribute'.  Setting it to true will result in a blank title attribute. 

If the pull request is accepted I'll update the wiki.


bbenezech:
Hi @superplussed 

An option to disable the title sounds too specific.
Can you add an option to change the title (and that of course would defaults to `strip_tags(value.to_s)`) and can be set to nil?
That would match the rest of the API a lot better.
With a test, I would happily merge it.

PR #2043, opened 2014-08-15
Fix time offset

nazarok:
The time offset is maintained if the type of field is Time


mshibuya:
Date/Time processing has been changed massively by #2391, so please reopen if you still need this change.

PR #2058, opened 2014-08-29
Be able to configure default association limit

quainjn:
Currently, you can configure association collection preloading on a per model basis.  If there's no custom configuration for a model, it will preload if the associated count is under 100.  This PR allows you to change that number on an application level.


bbenezech:
Perfect, thanks!

Issue #2047, opened 2014-08-19
Rails Admin with Bootstrap 3


remino:
Is there actual work done on getting Rails Admin to _use_ Bootstrap 3 itself instead of tucking Bootstrap 2 statically within its own gem?

I'd like to be able to create a layout for my app using Bootstrap 3 to reuse on both the public side and the admin side.

PR #2059, opened 2014-09-01
Moved to bootstrap 3

moskvin:
Related #2047


Jpunt:
Is the styling that I get with a fresh install correct? Seems like a big step backwards:

screen shot 2015-01-14 at 15 25 29

- Positioning of avatar
- Color of logout-button
- Colors of sidebar
- Right margins of list-table
- Smaller screens get something like this:

screen shot 2015-01-14 at 15 26 42

Is there something wrong with my setup? Or is this still a work in progress? In that case: is there a way to get back to the old theme?

PR #2085, opened 2014-10-14
Move sidebar navigation out into partial so it can be easily overridden

jordanmaguire:
This provides an easy way of overriding the existing sidebar navigation if you should choose to.

Just add a file `app/views/layouts/rails_admin/_sidebar_navigation.html.haml`

You can use a helper like below to create compatible `li`s:

```ruby
  # RailsAdmin expects a data-model attribute that it uses to determine whether or not this
  # li is selected. "data-model" must be a underscored, downcased version of the data_model
  # EG: For a class named "HospitalSupervisor", it will expect "hospital_supervisor"
  def rails_admin_sidebar_navigation_for(data_model, link_name=nil)
    # "HospitalSupervisor" becomes "Hospital Supervisors"
    link_name ||= data_model.titleize.pluralize
    content_tag(:li, "data-model" => data_model) do
      link_to(link_name, rails_admin.index_path(model_name: data_model.titleize.gsub(" ", "")))
    end
  end
```

Usage:

```ruby
rails_admin_sidebar_navigation_for("my_model_name")
# or
rails_admin_sidebar_navigation_for("my_model_name", "But display in sidebar like this")


jordanmaguire:
I've updated this branch so it is mergeable once more.

Appreciate if you could consider merging this bad boy.

mshibuya:
Thanks :+1:

Issue #1338, opened 2012-09-27
Nested polymorphic


Tuckie:
I'm attempting to have a nested polymorphic has_many with:

```ruby
class Promotion < ActiveRecord::Base
  belongs_to :promotable, :polymorphic => true, :inverse_of => :promotions
  attr_accessible :promotable_id, :promotable_type 
end
class Page < ActiveRecord::Base
  has_many :promotions, :as => :promotable, :inverse_of => :promotable, :dependent => :destroy
  accepts_nested_attributes_for :promotions, :allow_destroy => true
  attr_accessible :promotions_attributes, :allow_destroy => true
end
class Offering < ActiveRecord::Base
  has_many :promotions, :as => :promotable, :inverse_of => :promotable, :dependent => :destroy
  accepts_nested_attributes_for :promotions, :allow_destroy => true
  attr_accessible :promotions_attributes, :allow_destroy => true
end
```

I would like to have a nested form within each page or offering, but I'm getting
```
undefined method `abstract_model' for #<Array:0x00000004c6e130>
...
/usr/local/rvm/gems/ruby-1.9.3-p0/bundler/gems/rails_admin-f20879f82ea0/app/helpers/rails_admin/form_builder.rb:47:in `field_wrapper_for'
/usr/local/rvm/gems/ruby-1.9.3-p0/bundler/gems/rails_admin-f20879f82ea0/app/helpers/rails_admin/form_builder.rb:37:in `block (2 levels) in fieldset_for'
/usr/local/rvm/gems/ruby-1.9.3-p0/bundler/gems/rails_admin-f20879f82ea0/app/helpers/rails_admin/form_builder.rb:37:in `map'
/usr/local/rvm/gems/ruby-1.9.3-p0/bundler/gems/rails_admin-f20879f82ea0/app/helpers/rails_admin/form_builder.rb:37:in `block in fieldset_for'
actionpack (3.2.3) lib/action_view/helpers/capture_helper.rb:40:in `block in capture'
actionpack (3.2.3) lib/action_view/helpers/capture_helper.rb:187:in `with_output_buffer'
haml (3.1.7) lib/haml/helpers/xss_mods.rb:109:in `with_output_buffer_with_haml_xss'
actionpack (3.2.3) lib/action_view/helpers/capture_helper.rb:40:in `capture'
haml (3.1.7) lib/haml/helpers/action_view_mods.rb:105:in `capture_with_haml'
[...]

PR #2110, opened 2014-11-15
Fix for #1338

pencilcheck:
field.associated_model_config returns an array, but abstract_model is in field, this fixes the error


bbenezech:
That should work. Thanks!

PR #2133, opened 2014-12-10
Capitalize first letter of page titles and flash messages

suruja:
Hello!

This is a tiny pull request to display page titles and flash messages using the `capitalize_first_letter` helper.


mshibuya:
Just setting the corresponding wording to capitalized one will enough, I guess...

PR #2146, opened 2014-12-30
Ensure inline lists uses the right Bootstrap 3 css class

thechrisoshow:
According to the bootstrap migration readme http://getbootstrap.com/migration/ - the .inline class is deprecated on inline lists in favour of .list-inline.

The upshot of this for rails_admin is that the action icons on the index and dashboard page currently look 'squished' together.

This commit adds the .list-inline class  but also keeps the .inline class around for backwards compatibility. 

I notice that this is a common pattern in rails_admin for moving to Bootstrap 3, as there are several deprecated css classes dotted around the place. At some point these should be removed (moving to Bootstrap 3 *is* a breaking change)

PS: Travis-CI is reporting this as a broken build - but this appears to be GC limits with JRuby that's causing the error - unrelated to this pull request.

### DIFF
--- a/app/assets/stylesheets/rails_admin/base/theming.scss
+++ b/app/assets/stylesheets/rails_admin/base/theming.scss
@@ -269,7 +269,7 @@ body.rails_admin {
     /* icons */
     td.links {
       max-width: none;
-      .inline {
+      .list-inline {
         margin:0px;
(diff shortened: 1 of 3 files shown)


mshibuya:
Thanks! :+1:

PR #2148, opened 2015-01-02
The datepicker zindex needs to be higher for modals

thechrisoshow:
Currently if you click on a datepicker in a modal it is hidden behind the modal overlay. This commit brings it to the front.


mshibuya:
Thanks!

PR #2152, opened 2015-01-06
Fix sidebar-nav z-index issue when total_columns_width is large.

kayzee:
With rails_admin + Bootstrap 2, setting the total_columns_width to something arbitrarily large (to show all columns) works fine. The window scrolls to the right, showing a large data table.
However, doing the same thing in the latest rails_admin + Bootstrap 3 causes overlap with the sidebar-nav. See the image below.

Added z-index=100 to the .sidebar-nav class should solve the problem.

rails_admin-issue


mshibuya:
Thanks!

PR #2167, opened 2015-01-16
Add new item to the end of ordered multiselect instead of beginning

dwbutler:
I'm using the filtering multiselect in a nested form with `orderable: true`. For our use case, it is very important that elements are always in the right order.

I noticed that when you create a new item, visually the new `<option>` appears to be added as the last child of the `<select>`. But when the form is submitted to Rails, the new IDs have actually been inserted at the beginning instead of at the end.

I finally tracked this down to a hidden `<select>` which is the real one submitted via Ajax. Changing `prepend` to `append` fixed the issue, causing new `<option>`s to be inserted at the end of the hidden `<select>`.

I would be happy to write some specs, but I couldn't find any existing specs for the multiselect widget. Maybe I just need to be pointed to the right file or create a new one.


luizpicolo:
:+1:

dwbutler:
@sferik Anything blocking this from getting merged?

mshibuya:
Merging in, thanks!

dwbutler:
@mshibuya Awesome, thanks!

PR #2170, opened 2015-01-20
Set required attribute in the html elements

krishan1390:
Set required attribute in the html field and text elements based on the required flag on the field

### DIFF
--- a/app/views/rails_admin/main/_form_field.html.haml
+++ b/app/views/rails_admin/main/_form_field.html.haml
@@ -1,1 +1,1 @@
-= form.send field.view_helper, field.method_name, field.html_attributes.reverse_merge({ value: field.form_value, checked: field.form_value.in?([true, '1']), class: 'form-control'})
+= form.send field.view_helper, field.method_name, field.html_attributes.reverse_merge({ value: field.form_value, checked: field.form_value.in?([true, '1']), class: 'form-control', required: field.required})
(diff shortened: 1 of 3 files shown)


mshibuya:
Spec needed!

krishan1390:
We can set the required flag for a field in a model using:

    rails_admin do
      edit do
        field :name do
          required :true
        end
      end
    end

But this doesn't set the required attribute for the corresponding html input fields.

image


All it does is create a required label below the input field.
image

There is no client side validation and hence the form can be submitted without any value.

This fix is to set the required attribute for the text fields based on the above configuration.

mshibuya:
Oh, I'm sorry I wasn't clear.
We should have some RSpec test examples to ensure this change does not break in future.

I think they should be written here:
https://github.com/sferik/rails_admin/blob/master/spec/integration/basic/edit/rails_admin_basic_edit_spec.rb

krishan1390:
Done

mshibuya:
Thanks! :+1:

krishan1390:
Can you please take a look at https://github.com/sferik/rails_admin/issues/2171

Basically I also need it for the form_filtering_select and the form_filtering_multiselect partials

If I write this:

    = form.select field.method_name, collection, { selected: selected_id, include_blank: true }, field.html_attributes.reverse_merge({ data: { filteringselect: true, options: js_data.to_json }, placeholder: t('admin.misc.search'), required: field.required })

The required attribute is set to the "select" element. But this is invalid because it needs to be set to the autocomplete "input" element.

PR #2172, opened 2015-01-20
Adds POST as a verb for the delete action

matstc:
I was trying to delete a resource and received an error saying: `No route matches [POST] /author/1/delete`

The error was happening because the DELETE route for rails admin is expected to use either GET (to access the delete view) or DELETE (for the destroy action). But the form was generated to have its method attribute equal to POST. AFAICT, this is normal behavior and the real http verb is kept in a hidden field with name `_method`. See actionview-4.1.7/lib/action_view/helpers/form_tag_helper.rb @ line 714 ActionView::Helpers::FormTagHelper#extra_tags_for_form.

In any case, I made the changes in this PR and it is now working for me (although the flash message is not showing). Also I had to prefix `t` with `I18n`.

Am I missing something? Can anyone else delete resources? When you do, does the form submit a POST or a DELETE request?

I had similar problems with the edit action (also fixed by this PR).

### DIFF
--- a/app/controllers/rails_admin/main_controller.rb
+++ b/app/controllers/rails_admin/main_controller.rb
@@ -77,7 +77,7 @@ module RailsAdmin
     end
 
     def redirect_to_on_success
-      notice = t('admin.flash.successful', name: @model_config.label, action: t("admin.actions.#{@action.key}.done"))
+      notice = I18n.t('admin.flash.successful', name: @model_config.label, action: I18n.t("admin.actions.#{@action.key}.done"))
       if params[:_add_another]
         redirect_to new_path(return_to: params[:return_to]), flash: {success: notice}
       elsif params[:_add_edit]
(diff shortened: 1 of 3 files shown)


matstc:
FYI—I have similar problems when editing a resource.

mshibuya:
I don't get it.
What you say is partially true(with DELETE request, form's method becomes POST and real http verb is kept in `_method`), but [Rack::MethodOverride](https://github.com/rack/rack/blob/1.6.0/lib/rack/methodoverride.rb) automatically uses preserved `_method` value as HTTP method, so no special care in app will be needed.

Can you create minimal app to reproduce this problem?

matstc:
Thanks for the tip @mshibuya. I was using rails-api and the MethodOverride middleware wasn't being added to the stack.

I set `config.api_only = false` in `application.rb` and now the request goes through. However, I had to prefix all calls to `t` with `I18n`. Otherwise Rails can't find the method `t`. This is the only change in the PR. Is this change needed?

mshibuya:
By `parent_controller` config option(#2217), setting `RailsAdmin::ApplicationController`'s base class to `ActionController::Base` will solve this problem.

amoludage:
@matstc +1

PR #2182, opened 2015-02-03
exclude model action

### LINKED ISSUE #2023: Inconsistent behavior excluding models from actions
Hi, when I exclude a model from new action RailsAdmin should disallow its creation in any way across the whole admin interface.

RailsAdmin instead allows to create excluded models through association fields (it lets you open a modal fill all fields and save it) and just raises an un-handled error for breadcrumbs when you hit "save and add another" button.

This is pretty annoying and while you can disable 'inline_add' for associations it forces you to resort to authorization to hide the 'add_another' button.

The worst part is that controller actions aren't disabled: I can point the browser to new action path for a model and it always responds with the form. I've not tried but that makes me guess that I can destroy a model excluded from 'delete' action as well.

I think this should be fixed or documented someway.

Anyway I hate having to setup authorization for tasks like hiding a button. Especially if authorization Is not an app requirement (i.e nobody should be able to create a model no matter who he is).

The ideal solution would be to just disallow routes for disabled actions and change the `authorized?`  helper to check also if @abstract_model class is enabled for that action or not.

Something like:

``` rb
def authorized?(action_name, model)
   action_object = RailsAdmin::Config::Action.all.find {|a| a.action_name.to_s == action.to_s}
   if action_object && action_object.exclude.include?(model.model.name)
     return false
   else
     super(action, model)
   end
end
```

Thank you very much anyway,
if you wish to fix this I can look at the code deeper and help with a PR. 
Just let me know.

gagalago:
fix #2023

### DIFF
--- a/app/controllers/rails_admin/main_controller.rb
+++ b/app/controllers/rails_admin/main_controller.rb
@@ -9,6 +9,7 @@ module RailsAdmin
     before_filter :get_model, except: RailsAdmin::Config::Actions.all(:root).collect(&:action_name)
     before_filter :get_object, only: RailsAdmin::Config::Actions.all(:member).collect(&:action_name)
     before_filter :check_for_cancel
+    before_filter :prevent_disabled_actions, except: RailsAdmin::Config::Actions.all(:root).collect(&:action_name)
 
     RailsAdmin::Config::Actions.all.each do |action|
       class_eval %{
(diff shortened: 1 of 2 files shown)


mshibuya:
I've fixed it by 10655cbf, thanks anyway!

PR #2187, opened 2015-02-14
try to grab the current_user id first for paper trail

corbt:
With the current version of rails_admin and paper_trail 4.0.0.beta (not sure about earlier versions), the `whodunnit` field for updated versions just stores a ruby object reference -- something like `#<User:0x007f3434b429a8>` -- instead of an actual ActiveRecord id.

This change fixes the problem by explicitly attempting to pull an id out of the `current_user` first, and falling back to the default serialization if that doesn't work. It mirrors the paper_trail recommended technique discussed at https://github.com/airblade/paper_trail/issues/316.

### DIFF
--- a/app/controllers/rails_admin/application_controller.rb
+++ b/app/controllers/rails_admin/application_controller.rb
@@ -58,7 +58,9 @@ module RailsAdmin
       instance_eval(&RailsAdmin::Config.audit_with)
     end
 
-    alias_method :user_for_paper_trail, :_current_user
+    def user_for_paper_trail
+      _current_user.try(:id) || _current_user
+    end
 
     rescue_from RailsAdmin::ObjectNotFound do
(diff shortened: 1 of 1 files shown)


rikkipitt:
I can confirm that this works :+1:

batter:
You shouldn't need either of these... I would expect this to just work with the removal of the `alias_method :user_for_paper_trail, :_current_user`... perhaps you might need to modify it to `alias_method :current_user, :_current_user` due to the way [PaperTrail looks for an implementation of the `current_user` method](https://github.com/airblade/paper_trail/blob/v4.0.0.beta2/lib/paper_trail/frameworks/rails/controller.rb#L19) and then attempts to invoke the id attribute of the object if it is defined.

mshibuya:
Thanks!

batter:
:+1:

PR #2188, opened 2015-02-17
Updated _submit_buttons.html.haml. Fixes #2181.

### LINKED ISSUE #2181: Modal Has Two Sets of Buttons
image

I believe that this line is supposed to remove the duplicate set of buttons: https://github.com/sferik/rails_admin/blob/master/app/assets/javascripts/rails_admin/ra.remote-form.js#L73

But https://github.com/sferik/rails_admin/commit/f6d660fae06e97852e61422520e18fb9064f51bd changed `.form-actions` to `.form-group`, which probably should not be removed quite so aggressively.

rikkipitt:
Fixes #2181


mshibuya:
Thanks :+1:

PR #2198, opened 2015-02-25
Don't .capitalize the Home link name, as it removes capitalization

soupmatt:
Before this commit
admin.home.name = 'My Cool App' would be rendered 'My cool app'

After this commit
admin.home.name = 'My Cool App' would be rendered 'My Cool App'


mshibuya:
Thanks!

PR #2207, opened 2015-03-07
fix target typo 'blank' to '_blank'

kuboon:


lib/rails_admin/config/fields/types/file_upload.rb
```diff
@@ -35,7 +35,7 @@ class FileUpload < RailsAdmin::Config::Fields::Base
                 thumb_url = resource_url(thumb_method)
                 url != thumb_url ? v.link_to(v.image_tag(thumb_url, class: 'img-polaroid'), url, target: 'blank') : v.image_tag(thumb_url)
```
mshibuya: There's another `target: 'blank'` here...

kuboon: thanks. fixed.

mshibuya:
Thanks!

PR #2208, opened 2015-03-09
Fix dashboard history authorized check

rikkipitt:
When using CanCan (or CanCanCan) the ```can :history, :all``` ability doesn't show the history index on the dashboard. This minor PR fixes this issue by checking that the ```:history_index``` action rather than ```:history``` in app/views/rails_admin/main/dashboard.html.haml.


mshibuya:
Am sorry... this was side-effect of my change in 10655cbf.
 Thanks for your fix!

rikkipitt:
Thanks @mshibuya!

PR #2216, opened 2015-03-13
Change view of focused item in select dropdown

Daniel254:
Change view of focused item in select dropdown from http://prntscr.com/6gehla to http://prntscr.com/6geg2z
Prevents mooving text when focus in or focus out


mshibuya:
Thanks!

PR #2224, opened 2015-03-23
Fix Postgres column names

anfinil:
Rails Admin sometimes fails because column names are not quoted.


mshibuya:
Please fix [Travis errors](https://travis-ci.org/sferik/rails_admin/builds/55480924) and add specs for this :)

mshibuya:
Please reopen on update.

PR #2262, opened 2015-04-19
Fix form nested one help style

tagliala:
Model:
```rb
  has_one :country
  accepts_nested_attributes_for :country, allow_destroy: true, reject_if: :all_blank
```

### Before:
screen shot 2015-04-19 at 19 40 00

### After:
screen shot 2015-04-19 at 19 40 08

(as per: https://github.com/sferik/rails_admin/blob/master/app/views/rails_admin/main/_form_nested_many.html.haml#L1)


mshibuya:
Thanks!

PR #2263, opened 2015-04-20
Fix incorrect counts for capsuled models with the same name

DarthMax:
If I have 2 models that share the same name but are encapsulated in different modules the entity counts on the dashboard page are incorrect. This happens as the counts are stored in a hash using the value `AbstractModel#pretty_name` as key. This method itself uses `Model#model_name` which strips the module name. In this specific case this results in the same key for different models and therefore in wrong counts. 

This can be fixed by using `Model#name` instead of `Model#model_name` for the key.

### DIFF
--- a/spec/dummy_app/db/migrate/20150421143020_add_type_to_users.rb
+++ b/spec/dummy_app/db/migrate/20150421143020_add_type_to_users.rb
@@ -1,0 +1,6 @@
+class AddTypeToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :type, :string
+  end
+
+end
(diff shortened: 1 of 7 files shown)


mshibuya:
Write tests, please!

DarthMax:
Sorry. Here they are :)

spec/dummy_app/db/migrate/20150421143020_add_type_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddTypeToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :type, :string
+  end
+
+end
```
mshibuya: This migration does not affect test result at all, so should be removed.

mshibuya:
Please also create `User::Confirmed` model for [mongoid](https://github.com/sferik/rails_admin/tree/master/spec/dummy_app/app/mongoid).

mshibuya:
Looks okay, thank you!

PR #2264, opened 2015-04-21
Avoid to delete all with deleted resource ids

akm:
In the following cases, bulk_delete action deletes all records. This pull request fixes them.

## A. When delete the selected records before show confirmation

1. UserA visits list page
2. UserA selects records
3. UserB deletes the records which are selected by UserA
4. UserA clicks "Delete selected Functions Workers"
5. rails_admin renders confirmation form without buld_ids parameter
6. UserA clicks "Yes, I'm sure"
7. rails_admin deletes the resource without bulk_ids
    - All of the record in the table are deleted

## B. When delete the selected records before send DELETE request

1. UserA visits list page
2. UserA selects records
3. UserA clicks "Delete selected Functions Workers"
4. UserB deletes the records which are selected by UserA
5. UserA clicks "Yes, I'm sure"
6. rails_admin deletes the resource without bulk_ids
    - All of the record in the table are deleted

## C. When send DELETE request without bulk_ids

1. UserA sends the request without bulk_ids
    - by attacker
    - a mistake not from browser but using some HTTP client
2. rails_admin deletes the resource without bulk_ids
    - All of the record in the table are deleted

### DIFF
--- a/lib/rails_admin/config/actions/bulk_delete.rb
+++ b/lib/rails_admin/config/actions/bulk_delete.rb
@@ -15,10 +15,14 @@ module RailsAdmin
         register_instance_option :controller do
           proc do
             if request.post? # BULK DELETE
-
               @objects = list_entries(@model_config, :destroy)
 
-              render @action.template_name
+              if @objects.blank?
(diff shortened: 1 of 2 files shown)


mshibuya:
This is a terrible bug, thanks a lot!

PR #2270, opened 2015-04-24
Accept MySQL's 'utf8mb4' as a kind of UTF-8 for CSV export

dburt:
MySQL's default UTF-8 encoding is limited to 3-byte characters and
the encoding it uses and reports for full 4-byte Unicode is called
'utf8mb4' but if the database reports this encoding, the CSV export
module would try to convert string encodings utf8mb4 -> utf8,
causing an error.
Obviously the former is not a valid encoding name in Ruby, but is
actually going to be UTF-8, so this change assumes it's already
UTF-8.

### DIFF
--- a/spec/rails_admin/support/csv_converter_spec.rb
+++ b/spec/rails_admin/support/csv_converter_spec.rb
@@ -32,10 +32,23 @@ describe RailsAdmin::CSVConverter do
 
     subject { RailsAdmin::CSVConverter.new(objects, schema).to_csv(encoding_to: encoding) }
 
+    context 'when encoding FROM MySQL utf8mb4' do
+      let(:encoding) { 'UTF-8' }  # default
+
+      it 'exports to UTF-8 with BOM' do
+        # MySQL connection may report its encoding as 'utf8mb4'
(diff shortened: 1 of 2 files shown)


spec/rails_admin/support/csv_converter_spec.rb
```diff
@@ -32,10 +32,23 @@
 
     subject { RailsAdmin::CSVConverter.new(objects, schema).to_csv(encoding_to: encoding) }
 
+    context 'when encoding FROM MySQL utf8mb4' do
```
mshibuya: Please mark this as `active_record: true`, since it is causing Mongoid builds to fail.

dburt: Thanks for pointing me in the right direction. I was trying something more complex to solve this problem (something like skip unless connection.adapter_name == 'mysql2' ... which still has an advantage or two).

I've added the attribute to the `it` below because I think it reads better (you can read the 'when ... it ...' without interruption) and Travis is checking my work again and marking some mongoid scenarios green as I write this.

Thanks for your quick response and please do let me know if there's any other way you'd like this pull request improved.

mshibuya:
Thanks!

PR #2271, opened 2015-04-24
Fix csv for utf8mb4

dmedvinsky:
On MySQL, in order to store Emoji in the database, one must use `utf8mb4` encoding. Yet, this encoding was not present in the list of `UTF8_ENCODINGS`, so the export fails with

    ArgumentError: unknown encoding name - utf8mb4


dmedvinsky:
The build has failed but as far as I can tell, this is unrelated to my changes.

mshibuya:
Closing in favor of #2270.

PR #2273, opened 2015-04-27
Fix nested associations style

tagliala:
Apparently `col-md-offset-2` was not the proper class to set.

### Before:
screen shot 2015-04-27 at 21 15 55

### After:
screen shot 2015-04-27 at 21 14 50

I'm going to make another PR to fix checkbox's style


tagliala:
Not so simple...

PR #2274, opened 2015-04-27
Fix checkbox style

tagliala:
Remove checked option from generic form field
Add boolean field partial

**Please take a deeper look at this PR** because I added a new partial and removed the `checked` option from the generic field. I also need to set some custom css to avoid another label

The error is due to the `form-control` class currently added to the checkbox input

### Before
screen shot 2015-04-27 at 21 43 04

### After
screen shot 2015-04-27 at 21 42 16


mshibuya:
Thanks!

PR #2275, opened 2015-04-27
Fix nested associations style

tagliala:
Apparently `col-md-offset-2` was not the proper class to set. As other fields, we need `col-sm-10` plus a `clearfix` class on the `tab-content`

### Before:
screen shot 2015-04-27 at 21 15 55

### After:
screen shot 2015-04-27 at 21 14 50

#### Another screenshot
screen shot 2015-04-27 at 22 24 14

### DIFF
--- a/app/views/rails_admin/main/_form_nested_many.html.haml
+++ b/app/views/rails_admin/main/_form_nested_many.html.haml
@@ -1,4 +1,4 @@
-.controls.col-md-offset-2{data: { nestedmany: true }}
+.controls.col-sm-10{data: { nestedmany: true }}
   .btn-group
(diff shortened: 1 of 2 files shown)


tagliala:
Failure not caused by PR
  https://travis-ci.org/sferik/rails_admin/jobs/60276248

```
Bundler::GemspecError: Could not read gem at /home/travis/build/sferik/rails_admin/vendor/bundle/ruby/2.2.0/cache/rest-client-1.8.0.gem. It may be corrupted.
Installing nokogiri 1.6.6.2
An error occurred while installing rest-client (1.8.0), and Bundler cannot
continue.
Make sure that `gem install rest-client -v '1.8.0'` succeeds before bundling.
The command "eval bundle install --without development --jobs=3 --retry=3 --path=${BUNDLE_PATH:-vendor/bundle}" failed. Retrying, 2 of 3.
```

tagliala:
Could you please restart the travis build?

PR #2276, opened 2015-04-27
Set minimum Sass precision to 8

tagliala:
Bootstrap requires a [minimum sass precision of 8](https://github.com/twbs/bootstrap-sass#sass-number-precision)

This is causing some issues on button height. In example, if you try to append buttons to form controls the result will be 1px wrong

### Before
screen shot 2015-04-27 at 22 41 32

### After
screen shot 2015-04-27 at 22 46 11

(I'm trying to update the dropdown, but it seems I opened Pandora's box :grin:)

### DIFF
--- a/lib/rails_admin/bootstrap-sass.rb
+++ b/lib/rails_admin/bootstrap-sass.rb
@@ -19,6 +19,7 @@ module RailsAdmin
 
       stylesheets = File.expand_path(File.join('..', 'vendor', 'assets', 'stylesheets'))
       ::Sass.load_paths << stylesheets
+      ::Sass::Script::Value::Number.precision = [8, ::Sass::Script::Value::Number.precision].max
     end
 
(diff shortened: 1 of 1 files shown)


tagliala:
ping.

Here it is a screenshot showing the issue with the default precision of 5:
screen shot 2015-04-30 at 18 41 55

tagliala:
This could break things for some users

mshibuya:
My concern is that changing Sass precision globally might affect appearance of non-admin part in app.
Any thoughts on it?

tagliala:
@mshibuya Yes, I agree. This should be documented

tagliala:
Refers to https://github.com/sass/sass/issues/1122

tagliala:
Since this could impact on user application, I'm going to close this request and I will open an issue instead

PR #2284, opened 2015-05-01
Flash bootstrap alert styles

dburt:
This patch allows `redirect_to foo_path, notice: 'bar'` to show a coloured alert box rather than plain text.

Rails has some standard and commonly used flash keys, notice and alert, and some shortcuts to interact with them. But if you use them in Rails Admin, as in [this custom action tutorial](http://blog.endpoint.com/2012/03/railsadmin-custom-action-case-study.html), you'll need to add custom CSS for the `alert-notice` class, or else you must replace the standard Rails keys with the alert names from Bootstrap: `flash: {info: '...'}` instead of just `notice: '...'`

In the same way as Rails Admin already mapped the flash key 'error' to 'alert-danger' to give a red box, this patch additionally maps 'notice' to 'alert-info' for a blue box, and 'alert' to 'alert-warning' for a yellow box.

### DIFF
--- a/app/controllers/rails_admin/main_controller.rb
+++ b/app/controllers/rails_admin/main_controller.rb
@@ -115,7 +115,7 @@ module RailsAdmin
 
     def check_for_cancel
       return unless params[:_continue] || (params[:bulk_action] && !params[:bulk_ids])
-      redirect_to(back_or_index, flash: {info: t('admin.flash.noaction')})
+      redirect_to(back_or_index, notice: t('admin.flash.noaction'))
     end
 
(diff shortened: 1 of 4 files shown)


dburt:
Regarding the RuboCop failure on Travis: RuboCop is much faster than Rspec, so these failures could be detected much faster by swapping the order of the tasks in `Rakefile` like so:

```
task default: [:rubocop, :spec]
```

tagliala:
:+1:

mshibuya:
Merged in, thanks!

PR #2292, opened 2015-05-08
Fields with same names but new types should be placed in the original position

aquajach:
Fields, with the same names but different types shouldn't always be put at the bottom. 

For example, I have a couple fields in the middle of my form, :state and :state_updated_at as a pair of related fields. :state has a type as :string by default, but I would like to make it as :enum. Configuring the :state field will separate two fields and make :state appear at the bottom.


mshibuya:
Spec needed!

mshibuya:
Closed by #2409.

PR #2297, opened 2015-05-15
Clarify test description

beechnut:
Test used to say 'orders ... records by desc' (which is a little ambiguous).


mshibuya:
Thanks!

PR #2299, opened 2015-05-19
fix wysihtml5 asset path

gagalago:
fix path for use in production with rails 4.2 and last spockets
```
Asset was linked to from an alias rather than its exact path. Alias resolving may not be available in production. Use "bootstrap-wysihtml5/index.css" instead of "bootstrap-wysihtml5.css"
Asset was linked to from an alias rather than its exact path. Alias resolving may not be available in production. Use "bootstrap-wysihtml5/index.js" instead of "bootstrap-wysihtml5.js"
```

### DIFF
--- a/lib/rails_admin/config/fields/types/wysihtml5.rb
+++ b/lib/rails_admin/config/fields/types/wysihtml5.rb
@@ -16,11 +16,11 @@ module RailsAdmin
           end
 
           register_instance_option :css_location do
-            ActionController::Base.helpers.asset_path('bootstrap-wysihtml5.css')
+            ActionController::Base.helpers.asset_path('bootstrap-wysihtml5/index.css')
           end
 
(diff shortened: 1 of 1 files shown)


gagalago:
This pull request don't fix the wysihtml5 in production. That works in development but not in production.

AlexEscalante:
I updated my app to Rails 4.1.2 and Ruby 2.2.2 and I am having this very same issue. It started with:

    Asset filtered out and will not be served: add `Rails.application.config.assets.precompile += %w( bootstrap-wysihtml5/index.css )` to `config/initializers/assets.rb` and restart your server

Following the advice didin't work, I got the message shown on entries above...

Any ideas?

AlexEscalante:
I just tried these same changes and they work both on development and production, tried them on Heroku... I will submit the pull request in case someone wants to check it out…

PR #2303, opened 2015-05-22
Fix for Image dropdown not working

GeoffTidey:
tracked down to a change in b05349838e0ea074ec0bdd49e785e63c0372e81d using git bisect :)


mshibuya:
Thanks!

PR #2305, opened 2015-05-22
allow to retrieve a scoped index from a polymorphic lazyloaded association

zambot:
in the view, we have two dropdowns to select type and object in a polymorphic belongs_to association.
with these changes you have the ability to configure a scope on the association instead of always retrieving all objects of the selected type


mshibuya:
Closing in favor of #2886.

PR #2306, opened 2015-05-23
Fixing sprockets wysihtml5 problem

AlexEscalante:
This solution was attempted on pull request #2299 by @gagalago. He mentions it didn't work on production, however I find it does quite well. Could you check it out?


AlexEscalante:
My bad. I should have tested more. Styles don't seem to be applied correctly on production. We need a better fix!

AlexEscalante:
I had a typo, sorry :( I will reissue…

PR #2307, opened 2015-05-23
Fixing sprockets wysihtml5 problem (2)

AlexEscalante:
This works but labels are missing from some toolbar buttons. I guess this is another issue to look into.

### DIFF
--- a/lib/rails_admin/config/fields/types/wysihtml5.rb
+++ b/lib/rails_admin/config/fields/types/wysihtml5.rb
@@ -16,7 +16,7 @@ module RailsAdmin
           end
 
           register_instance_option :css_location do
-            ActionController::Base.helpers.asset_path('bootstrap-wysihtml5/index..css')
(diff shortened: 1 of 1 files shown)


luizpicolo:
I'm experiencing this problem as well. Good job. :+1:

mshibuya:
Asset alias check has been reverted by rails/sprockets-rails#251.
Using the master of sprockets-rails works now.
```ruby
gem 'sprockets-rails', github: 'rails/sprockets-rails'
```

luizpicolo:
Thats good. resolved my problem. Thanks :+1:

vorant94:
Does anybody have another solution? Because It crashes with using compass-rails gem
````
Bundler could not find compatible versions for gem "sprockets":
  In Gemfile:
    compass-rails (>= 0) ruby depends on
      sprockets (< 2.13) ruby

    sprockets-rails (>= 0) ruby depends on
      sprockets (3.0.0)
````

PR #2309, opened 2015-05-27
Fixed issue with nav does not check pjax config from an action.

fanminjian:
Hopefully this will be a quick fix for that issue.


tagliala:
Failures due to #2311

mshibuya:
Test needed!

bbenezech:
Good catch!
Thanks @tagliala 
@mshibuya I don't think a test is needed here.

PR #2313, opened 2015-05-28
Fix delete button for nested form

dalpo:
#### Before:

screenshot from 2015-05-28 22 16 37

#### After:

screenshot from 2015-05-28 22 14 05


mshibuya:
Thanks!

PR #2315, opened 2015-05-29
Fix date and datetime input fields markup

dalpo:
#### Before
screenshot from 2015-05-29 20 21 53

#### After
screenshot from 2015-05-29 20 18 58


rikkipitt:
Much nicer!

mshibuya:
Thank you!

PR #2319, opened 2015-06-01
Fix nested form controls

herk:
As of #2313, you cannot switch between multiple objects as the content tab is above the controls.
rails_admin_nested_form
So in the form above I cannot change to the previous tab (unless I delete the active one)


luizpicolo:
Very good. Fixed my problem. Tanks :+1:

mshibuya:
It was pretty annoying, thanks!

PR #2323, opened 2015-06-09
Included missing Bootstrap panels scss and added missing bottom margi…

### LINKED ISSUE #2282: Bootstrap panels
I'm trying to use Bootstrap panel http://getbootstrap.com/components/#panels in a custom view for a custom action, but the style is not applying. I can use a table style http://getbootstrap.com/css/#tables so I assume that bootstrap is included and I can use it. 
Do I need another configuration to use panels?

panels
table

rikkipitt:
…n to the nav tabs

Will fix #2282 and makes the filter "well" under the nav tabs sit like they used to in earlier versions of rails_admin (i.e. not touching).

Before:

before

After:

after

### DIFF
--- a/app/assets/stylesheets/rails_admin/base/theming.scss
+++ b/app/assets/stylesheets/rails_admin/base/theming.scss
@@ -191,8 +191,12 @@ body.rails_admin {
   }
 
   /* tab links should match icon color */
-  .nav.nav-tabs li.icon a {
-    color: #000;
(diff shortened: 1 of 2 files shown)


rikkipitt:
I don't think those Travis errors are related to this pull request...

dalpo:
:+1:

mshibuya:
Thanks :smiley:

PR #2331, opened 2015-06-15
Do not monkey patch the app's YAML

nwjsmith:
SafeYAML provides a way to avoid monkey-patching the `YAML` object. This
should be used to avoid overwriting the application's `YAML`.

This was causing problems with our application, which uses Psych's `YAML.safe_load` method in ways that are incompatible with SafeYAML's implementation. It's important to note that `YAML.safe_load` has existed in Ruby's stdlib [since 2.1.0][1].

[1]: http://ruby-doc.org/stdlib-2.1.0/libdoc/psych/rdoc/Psych.html#method-c-safe_load

### DIFF
--- a/lib/rails_admin/config/fields/types/serialized.rb
+++ b/lib/rails_admin/config/fields/types/serialized.rb
@@ -14,7 +14,7 @@ module RailsAdmin
 
           def parse_input(params)
             return unless params[name].is_a?(::String)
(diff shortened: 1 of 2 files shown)


mshibuya:
Thanks!

bf4:
I guess https://github.com/sferik/rails_admin/pull/2397 got lost?

PR #2335, opened 2015-06-19
Fix: HTML string inputs should not have a size attribute valorized with 0

dalpo:
Related to https://github.com/rails/rails/pull/14579


mshibuya:
Thanks :ok_hand: :

dalpo:
Thanks to you @mshibuya :smile:

PR #2339, opened 2015-06-20
Change: make browser validations configurable via DSL

dalpo:
Add the ability to enable/disable html5 browser validations through:

```ruby
RailsAdmin.config do |config|
  config.browser_validations = true
end
```

### DIFF
--- a/lib/rails_admin/config.rb
+++ b/lib/rails_admin/config.rb
@@ -50,7 +50,7 @@ module RailsAdmin
       # hide blank fields in show view if true
       attr_accessor :compact_show_view
 
-      #  Enable HTML5 browser validation
+      # Tell browsers whether to use the native HTML5 validations (novalidate form option).
       attr_accessor :browser_validations
 
       # Set the max width of columns in list view before a new set is created
--- a/lib/rails_admin/config.rb
+++ b/lib/rails_admin/config.rb
@@ -50,6 +50,9 @@ module RailsAdmin
       # hide blank fields in show view if true
       attr_accessor :compact_show_view
 
+      #  Enable HTML5 browser validation
+      attr_accessor :browser_validations
+
       # Set the max width of columns in list view before a new set is created
       attr_accessor :total_columns_width
 
@@ -254,6 +257,7 @@ module RailsAdmin
       # @see RailsAdmin::Config.registry
       def reset
         @compact_show_view = true
+        @browser_validations = true
         @yell_for_non_accessible_fields = true
         @authenticate = nil
         @authorize = nil
--- a/app/helpers/rails_admin/main_helper.rb
+++ b/app/helpers/rails_admin/main_helper.rb
@@ -3,7 +3,11 @@ require 'builder'
 module RailsAdmin
   module MainHelper
     def rails_admin_form_for(*args, &block)
-      options = args.extract_options!.reverse_merge(builder: RailsAdmin::FormBuilder)
+      options = args.extract_options!.reverse_merge(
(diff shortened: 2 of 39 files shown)


dalpo:
I'm not able to reproduce the failure with mongoid...

dalpo:
Actually, browser validations are generating a lot of javascript errors for has_many/has_one widget becouse the input fields are hidden.

For instance:

```
An invalid form control with name='' is not focusable.
An invalid form control with name='recipe[main_variant_attributes][variant_ingredients_attributes][1435052695601][quantity]' is not focusable.
An invalid form control with name='' is not focusable.
An invalid form control with name='recipe[main_variant_attributes][variant_ingredients_attributes][1435052695601][quantity]' is not focusable.
```

dalpo:
@mshibuya Let me know what do you think about this PR :wink:

andreazaupa:
+1

giacomomacri:
+1

lib/rails_admin/config.rb
```diff
@@ -254,6 +257,7 @@ def models
       # @see RailsAdmin::Config.registry
       def reset
         @compact_show_view = true
+        @browser_validations = false
```
mshibuya: I'm a bit concerned about defaulting this to `false`.
Don't you think that client-side validation is a nice-to-have feature?

dalpo: @mshibuya If you prefer I can set the default value to true. It's not a problem for me. 

My personal opinion is that browser validation is a nice to have feature only for simple application's models.

In my experience there are two use cases that it could be a trouble:

- When a model has has_many/has_one nested attributes to be validated the browser-validation is generating a lot of js errors as commented above. 

- When a model has conditional validations.


Let me know what do you prefer as default value so I can do a commit and rebase the PR :wink:

dalpo:
@mshibuya I changed the default value to `true`

dalpo:
@mshibuya @sferik any news about this PR?

mshibuya:
Squashed and cherry-picked with slight modification(`html_required?` is unnecessary, because setting `novalidate` to form is enough). Thanks!

dalpo:
:heart: 

Thanks @mshibuya!!

fmh:
@dalpo 
do you need that or is it an oversight? : 
`````ruby
register_instance_option :html_attributes do
          {
            required: required?,
          }
   end
``````
in my case , config.browser_validations  = false has no effect, and browser validation still works for me.

**Comment by dalpo** [comment]
@fmh Thanks for reporting!

I fixed it here: https://github.com/sferik/rails_admin/pull/2373

**Comment by fmh** [comment]
now, it works, thx 👍

**Comment by vincentwoo** [comment]
@dalpo as you said, I am getting a bunch of the following js errors on rails_admin 0.7:

```
An invalid form control with name='' is not focusable.
An invalid form control with name='recipe[main_variant_attributes][variant_ingredients_attributes][1435052695601][quantity]' is not focusable.
```

This did not happen on 0.6.8. Disabling all browser validation feels like overkill. Did we figure out what was wrong in this case?

dalpo:
@vincentwoo I didn't investigate too much about it.

Btw this was the old PR that introduced the required attribute: 
https://github.com/sferik/rails_admin/pull/2170/files

As you can see that PR didn't introduce the required attribute for all input fields, but only for those that are using the `_form_field.html.haml` and `_form_text.html.haml` partials.

However with the latest changes, now the required attribute is present for every input field.
May be this could cause your errors.
 
P.S.
Sorry for my bad English..

SuperMasterBlasterLaser:
I have found some bug on browser validation #2492

PR #2349, opened 2015-06-30
main_controller/get_collection : all associations should be searchable

voondo:
Related to #1434

### DIFF
--- a/app/controllers/rails_admin/main_controller.rb
+++ b/app/controllers/rails_admin/main_controller.rb
@@ -119,7 +119,7 @@ module RailsAdmin
     end
 
     def get_collection(model_config, scope, pagination)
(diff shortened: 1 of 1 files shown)


mshibuya:
We can't take this approach, because including `has_many` association causes massive amount of DB traffic(conversely, including `belongs_to` or `has_one` association is OK, because they are guaranteed to have  only 1 record).
We need some sort of configuration to decide inclusion of an association.

voondo:
@mshibuya IMHO it's the problem of the developers to manage performance issues by enabling or not  the searchable fields. Anyway this patch also fixes an issue of associations with a "through" (eg. has_one :stuff, through: :thing).
This issue is not an edge case and it seems many people are facing with it.
Maybe you should disable the searchable option by default (it is very simple to enable).

PR #2362, opened 2015-07-19
Fix tags rendered as text with jquery .text()

fmh:
after f15fd21 commit.

capture d ecran 2015-07-19 a 16 24 49

### DIFF
--- a/app/assets/javascripts/rails_admin/ra.filtering-select.js
+++ b/app/assets/javascripts/rails_admin/ra.filtering-select.js
@@ -92,9 +92,10 @@
         input.attr('placeholder', select.attr('placeholder'))
 
       input.data("ui-autocomplete")._renderItem = function(ul, item) {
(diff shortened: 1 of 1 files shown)


herk:
This seems a bit unnecessary to me, I'd remove the <strong> tag from https://github.com/sferik/rails_admin/blob/master/app/assets/javascripts/rails_admin/ra.filtering-select.js#L131 instead.

mshibuya:
This is a bad change, because it strips a label `x < 1 and y > 5` into `x  5`.

Please remove tags from your label. It was unintentional feature.

mshibuya:
I'm a noob, I just thought that the `strong` tags were something inserted via app's custom logic.
Already addressed by 7aee78bf so please take a look.

@fmh @herk Thanks you guys for using the master branch and help us make RailsAdmin better!

Issue #2366, opened 2015-07-28
Exclude model which has no db-related functionality


olegantonyan:
I'm using [active_type](https://github.com/makandra/active_type) or all sort of service/form classes which should quack like an activerecord. It has some advantages over `include ActiveModel::Model`.

But, its base class `ActiveType::Object` is basically `ActiveRecord::Base` with [all db-related features turned off](https://github.com/makandra/active_type/blob/master/lib/active_type/no_table.rb). Because it is `ActiveRecord::Base`, `rails_admin` tries to query db for all its records and fails with 'no relation' error. 

Sure, I can put all `ActiveType::Object` models into `excluded_models`, but I feel like there is a better approach: to have an ability to exclude a model with all its children, OR (maybe better) do not list models without db, for example if `columns` or `attribute_names` is empty or something like that.

bbenezech:
Thank you for reporting and providing a work around. (`excluded_models`)
I close it as won't fix because I don't think this is a much reported pb.

PR #2367, opened 2015-07-28
Exclude models without database (#2366)

olegantonyan:
This is a basic idea for issue #2366 
There are some troubles with tests. On a vanilla project even without any changes I've got 200-300 failures. I've no idea what's wrong with them.


bbenezech:
This is too specific, sorry.
RailsAdmin lists all ActiveRecord models by default.
I am afraid you should go the exclude_models way.

PR #2368, opened 2015-07-29
Provide the same styles for static link header as other nav headers

hube:
Before:
before

After:
after


mshibuya:
Closing in favor of #2378, thanks.

hube:
Great, thanks

PR #2374, opened 2015-08-09
Separate filter code and add option to make model filterable or not

fmh:
lighten index (list) file by separate filter code to a partial and add option to make model filterable or not


fmh:
must be two separate PR (1 for pjax and 1 for proposed "filter code" )

PR #2378, opened 2015-08-13
fix static_navigation style

josedonizetti:
static_navigation is using a different class of main_navigation, layout is broken if you have static links


mshibuya:
Thanks!

PR #2379, opened 2015-08-13
Add ability to view has_many associations

corbt:
I've found it generally useful to be able to view all the `has_many` associations associated with an object.

As a concrete example, in the app I'm working on a user can take video recordings, which are associated with their user account through a `has_many` association. Video files take a while to process after being uploaded, and it's helpful to be able to see the state of all of a user's recordings. Unfortunately, the current default implementation of rails_admin's `association` field type only shows the name of each association, and to find out more details you'd need to click on every associated object. Since in my case a user may have dozens of recordings, this becomes tedious quickly.

This pull request introduces a new `view_url` instance option that, when present in a rails_admin field, adds a link in the field's header on the show page. Additionally, I've created an implementation of this method for the `association` field type that, when more than one associated object is present (and the associated object is registered with rails_admin), leads to a filtered index page for the associated model showing the objects associated with the object in question.

If that didn't make sense, hopefully these screen shots for my `User` and `Recording` class make it clear how this works (the "View" link from the first image leads to the index page seen in the second).

screen shot 2015-08-13 at 7 50 10 pm

screen shot 2015-08-13 at 7 50 37 pm

### DIFF
--- a/app/assets/stylesheets/rails_admin/bootstrap/_labels.scss
+++ b/app/assets/stylesheets/rails_admin/bootstrap/_labels.scss
@@ -26,6 +26,11 @@
     position: relative;
     top: -1px;
   }
+
+  a {
+    color: $label-color;
+    text-decoration: underline;
+  }
 }
 
 // Add hover effects, but only for links
--- a/app/views/rails_admin/main/show.html.haml
+++ b/app/views/rails_admin/main/show.html.haml
@@ -12,5 +12,7 @@
               %dt
                 %span.label.label-info{class: "#{field.type_css_class} #{field.css_class}"}
                   = field.label
+                  - if field.try(:view_url)
+                    %a{href: field.view_url} (#{t('admin.actions.show.view')})
               %dd.well
                 = field.pretty_value
--- a/config/locales/rails_admin.en.yml
+++ b/config/locales/rails_admin.en.yml
@@ -73,6 +73,7 @@ en:
         title: "Details for %{model_label} '%{object_label}'"
         menu: "Show"
         breadcrumb: "%{object_label}"
+        view: "View"
       show_in_app:
         menu: "Show in app"
       new:
--- a/lib/rails_admin/config/fields/association.rb
+++ b/lib/rails_admin/config/fields/association.rb
@@ -14,6 +14,25 @@ module RailsAdmin
           @properties
         end
 
+        register_instance_option :view_url do
+          current_model = RailsAdmin.config(bindings[:object]).abstract_model.to_param
+          associated_model = associated_model_config.abstract_model.to_param
+
+          # Make sure the appropriate field is available on the associated model,
+          # and that there is more than one value available (if not this view
+          # isn't very helpful)
+          if associated_model_config.fields.map(&:name).include?(current_model.to_sym) &&
+              [value].flatten.size > 1
+            bindings[:view].url_for(
+              action: :index,
+              model_name: associated_model,
+              "f[#{current_model}][1318][o]" => "is",
+              "f[#{current_model}][1318][v]" => bindings[:object].id)
+          else
+            nil
+          end
+        end
+
         register_instance_option :pretty_value do
           v = bindings[:view]
           [value].flatten.select(&:present?).collect do |associated|


fmh:
I had the same need, but I proceeded otherwise.
1- create "_has_many_details_html.html" in rails_admin/main
``````ruby
:ruby
  require 'rails_admin/config/fields/types/datetime.rb'

  @model_config = model_config
  @abstract_model = abstract_model
  @objects = objects

  params = request.params.except(:authenticity_token, :action, :controller, :utf8, :bulk_export, :_pjax)

  params.delete(:sort_reverse) unless params[:sort_reverse] == 'true'

  params.delete(:sort) if params[:sort] == @model_config.list.sort_by.to_s

  properties = @model_config.list.with(controller: self.controller, view: self, object: @abstract_model.model.new).visible_fields
  # columns paginate
  @filterable_fields = @model_config.list.fields.select(&:filterable?)
  sets = get_column_sets(properties)
  properties = sets[params[:set].to_i] || []


#list
  %style

    - properties.select{ |p| p.column_width.present? }.each do |property|
      = "#list th.#{property.css_class} { width: #{property.column_width}px; min-width: #{property.column_width}px; }"
      = "#list td.#{property.css_class} { max-width: #{property.column_width}px; }"


  %table.table.table-condensed.table-striped.table-hover
    %thead
      %tr
        %th.shrink
        - properties.each do |property|
          %th= capitalize_first_letter(property.label)

    %tbody
      - @objects.each do |object|
        %tr{class: "#{@abstract_model.param_key}_row"}
          %td
            = link_to "<i class='btn btn-primary fa fa-info-circle'></i>".html_safe , show_path(model_name:abstract_model,id:object.id) , class:'pjax'
            = link_to "<i class='btn btn-info fa fa-pencil'></i>".html_safe , edit_path(model_name:abstract_model,id:object.id), class:'pjax'
          - properties.map{ |property| property.bind(:object, object) }.each do |property|
            - value =  property.pretty_value
            %td{class: "#{property.css_class} #{property.type_css_class}", title: strip_tags(value.to_s)}= value
`````
2- some code in show.html.haml , replace line 16
````ruby
 - if field.type == :has_many_association
                    = render partial: 'has_many_details', locals: {objects: @object.send(field.name) , model_config: field.associated_model_config , abstract_model: field.associated_model_config.abstract_model}
                  - else
                    = field.type == :ck_editor ? field.pretty_value.html_safe : field.pretty_value
````
Result : 

![capture d ecran 2015-08-22 a 02 55 37](https://cloud.githubusercontent.com/assets/56861/9421770/29e652b2-487a-11e5-8d99-82f7007e7cc1.png)

maximski:
@fmh would be pretty neat if Rails Admin would implement your solution (eventually you could add another option in the RailsAdmin.config for model list and show to enable the detailed view or not).

SuperMasterBlasterLaser:
@corbt I have used simple stupid solution:

       def self.items_as_html_table(items, bindings)
           table_tag = "<table class=\"table table-condensed table-striped\">"

           table_tag += '<thead><tr>'
           table_tag += "<th>First attr</th>"
           table_tag += "<th>Second attr</th>"
           table_tag += "<th>Third attr</th>"
           table_tag += "<th>Fourth attr</th>"
           table_tag += '</tr></thead>'

           table_tag += '<tbody>'

           items.each do |item|
              table_tag += '<tr>'
              path = bindings[:view].show_path(model_name: 'item', id: item.id.to_s)
              table_tag += "<td><a href=\"#{path}\">#{item.firsta></td>"
              table_tag += "<td>#{item.second}</td>"
              table_tag += "<td>#{item.third}</td>"
              table_tag += "<td>#{item.fourth}</td>"
              table_tag += '</tr>'
           end

         table_tag += '</tbody>'

         table_tag += '</table>'

         table_tag
    end

In `show` action I have shown it as `pretty_value` and with `html_safe` method.

I know that this is stupid solution. But for now it works. However, it will affect on performance when child models are more than 1000. I hope that your pull request will be accepted soon.

mshibuya:
Sorry for unresponsiveness, but I can't merge a feature without tests.

PR #2383, opened 2015-08-15
Fix removal buttons within filtringselect parents

### LINKED ISSUE #2372: nested many form not work in modal
this issue is just in modal 
before click to add nested form (nested many)
capture d ecran 2015-08-06 a 15 54 36

after click
capture d ecran 2015-08-06 a 15 55 01

it is caused by something that removes .btn-group content

fmh:
Only add and edit buttons are concerned.
Fix #2372


mshibuya:
Thanks!

PR #2393, opened 2015-08-28
adds redactor support

JenHomann:
Redactor requires a license. Users will need to put the redactor.js and redactor.css into their project manually.


anarchocurious:
+1

mshibuya:
Thanks for the request, but I don't like having proprietary-software integration in RA.
Please consider implementing as [RailsAdmin plugin](https://github.com/sferik/rails_admin/wiki#field-extension) instead.

PR #2401, opened 2015-09-15
Improve efficiency of filter query in Postgres

dwbutler:
In Postgres, filter queries currently generate a condition `LOWER(field) ILIKE ?`. This defeats the benefit of using ILIKE, which is case insensitive, and also prevents an index from being used. This fixes the issue by changing the condition to `field ILIKE ?` in Postgres.


mshibuya:
Thanks!

dwbutler:
Thanks for the quick response!

PR #2404, opened 2015-09-17
Import custom variables before Bootstrap variables.

johanneskrtek:
Take advantage of the Sass !default flag in order to configure Bootstrap.

### DIFF
--- a/app/assets/stylesheets/rails_admin/rails_admin.scss.erb
+++ b/app/assets/stylesheets/rails_admin/rails_admin.scss.erb
@@ -35,10 +35,10 @@
 
 /***  Variables  ***/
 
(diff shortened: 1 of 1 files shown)


mshibuya:
Merged in, thanks!
It would be great if you could add instruction to [Wiki](https://github.com/sferik/rails_admin/wiki/Theming-and-customization) on how to make use of this change :pray:

johanneskrtek:
Ok, I will add instructions to the wiki one of these days.

johanneskrtek:
Done. (I also updated the broken links to the bootstrap-sass repo.)

mshibuya:
Thanks :smiley:

PR #2409, opened 2015-09-23
Changing a field's type shouldn't position it at the bottom

aquajach:
Fields that have been re-configured with different types should remain at the original positions, not append at the end of the fields list

For example, I have a couple fields in the middle of my form, :state and :state_updated_at as a pair of related fields. :state has a type as :string by default, but I would like to make it as :enum. Configuring the :state field will separate two fields and make :state appear at the bottom. 

🙏🙏🙏


mshibuya:
Thanks!

Issue #1674, opened 2013-06-18
undefined method `attachment_definitions' for #<Class:0x007fa004a0c908>


tradyblix:
I have everything running on rails 4, but I'm having this issue with paperclip, it seems broken in rails admin rails-4 branch, wondering if anyone else knows what's going on or has seen this. Any paperclip model throws :

undefined method `attachment_definitions' for #<Class:0x007fa004a0c908>

This is with the latest version 3.4.2 of paperclip. 

Thanks

jemminger:
I'm using paperclip 4.2.1 and still getting this error.

mshibuya:
Can you paste full stack trace?

jemminger:
```

URL: https://mysite/rails_admin/ckeditor~asset

vendor/bundle/ruby/2.2.0/gems/activerecord-4.2.0/lib/active_record/dynamic_matchers.rb:26 in method_missing
vendor/bundle/ruby/2.2.0/gems/rails_admin-0.6.6/lib/rails_admin/config/fields/factories/paperclip.rb:8 in block in <top (required)>
vendor/bundle/ruby/2.2.0/gems/rails_admin-0.6.6/lib/rails_admin/config/fields.rb:54 in call
vendor/bundle/ruby/2.2.0/gems/rails_admin-0.6.6/lib/rails_admin/config/fields.rb:54 in block (2 levels) in factory
vendor/bundle/ruby/2.2.0/gems/rails_admin-0.6.6/lib/rails_admin/config/fields.rb:54 in each
vendor/bundle/ruby/2.2.0/gems/rails_admin-0.6.6/lib/rails_admin/config/fields.rb:54 in detect
vendor/bundle/ruby/2.2.0/gems/rails_admin-0.6.6/lib/rails_admin/config/fields.rb:54 in block in factory
vendor/bundle/ruby/2.2.0/gems/rails_admin-0.6.6/lib/rails_admin/config/fields.rb:50 in each
vendor/bundle/ruby/2.2.0/gems/rails_admin-0.6.6/lib/rails_admin/config/fields.rb:50 in factory
vendor/bundle/ruby/2.2.0/gems/rails_admin-0.6.6/lib/rails_admin/config/has_fields.rb:130 in _fields
vendor/bundle/ruby/2.2.0/gems/rails_admin-0.6.6/lib/rails_admin/config/has_fields.rb:133 in _fields
[...]

PR #2410, opened 2015-09-23
Fix error with paperclip

dmilisic:
Fix for http://stackoverflow.com/questions/17523849/rails-4-paperclip-and-rails-admin-undefined-method-attachment-definitions-er

Paperclip factory tries to register paperclip type even when there is no `has_attached_file` definition in the model. That causes `undefined method 'attachment_definitions'` erorr.


mshibuya:
Merging into #1674.

PR #2411, opened 2015-09-24
Add missing .alert-dismissible class to flash

johanneskrtek:
In Bootstrap the correct styling of the .close button (use a button element) depends on the .alert-dismissible class.


mshibuya:
Thanks!

PR #2414, opened 2015-10-01
Remove the field on "nested:fieldRemoved"

yewton:
* to avoid useless error on save when removed nested fields have some
  HTML5 validation errors


yewton:
This causes another problem that I cannot remove the existing nested record.
So I discard this.

PR #2417, opened 2015-10-01
Update button class to Bootstrap 3.

johanneskrtek:
Bootstrap 3 .btn class is not intented to be used without a modifier (i.e. btn-default).


mshibuya:
Thanks!

PR #2418, opened 2015-10-01
Space out caret in tab dropdown-toggle.

johanneskrtek:
Add missing whitespace.


mshibuya:
What kind of problem is this fix trying to solve?
If it is a appearance issue, could you provide screenshots?

johanneskrtek:
Yes it is an appearance issue; have a look at the attached screenshots (Chrome v45): on the left side the dropdown-toggle from the Bootstrap 3 documentation and on the right side the RailsAdmin implementation.

Imho it is bad practice to rely on whitespace for styling, but well, as Bootstrap does it this way, let's stick to it.

dropdown-toggles

mshibuya:
Merged in, thanks!

PR #2419, opened 2015-10-04
Show sidebar scrollbar only on demand.

johanneskrtek:
Avoid permanent visible scrollbar in sidebar by setting `overflow-y` to `auto` instead of `scroll`.


dalpo:
+1

mshibuya:
Thanks!

PR #2421, opened 2015-10-06
Change: add support for Array fields

dalpo:


mshibuya:
The name is too broad.
What's the difference with multiple-enabled enum?
https://github.com/sferik/rails_admin/blob/master/lib/rails_admin/config/fields/types/enum.rb#L33

OleksandrBezhan:
It seems like you didn't merge this to the master. I don't see array support in [config/fields/types/all.rb](https://github.com/sferik/rails_admin/blob/v0.8.1/lib/rails_admin/config/fields/types/all.rb)

Issue #1348, opened 2012-10-03
add "unscoped" mode to "model config"


ideomix:
I want to add a "unscoped" mode to "model config".
I also read the entry in the past, such a way as to create a subclass is difficult to accept.
To "model config" section of the "raid_admin.rb" simple, strongly desired a solution that can be configured as "unscoped = true".

What everyone think?

cassidyclawson:
+ 1 

Just another vote for this feature.

dmitrypol:
+1

mshibuya:
In fact, this can already be achieved by list scope configuration(added by #1861).
```ruby
RailsAdmin.config do |config|
  config.model Player do
    list do
      scopes [:unscoped]
    end
  end
end
```

mshibuya:
As is stated in https://github.com/sferik/rails_admin/pull/2428#issuecomment-147064737, above solution only works for list view.
I'll try to work this out...

dmitrypol:
When using RailsAdmin 0.7.0 with Mongoid (4 and 5) I get this error when accessing http://localhost:3000/admin/user?model_name=user&scope=unscoped

Here is my RA config:
    config.model 'User' do
      list do      
        scopes  [nil, :unscoped]
      end
    end

gems/ruby-2.0.0-p594/gems/rails_admin-0.7.0/app/views/rails_admin/main/index.html.haml where line #154 raised:

nil can't be coerced into Float
Extracted source (around line #26):
      count_without_padding = 0 if count_without_padding < 0

26      total_pages_count = (count_without_padding.to_f / limit_value).ceil
      if max_pages.present? && max_pages < total_pages_count
        max_pages
      else

dburt:
@mshibuya please let me know if you can use my help. For example, if my pull request is not too far off functionally, you could show me what you want the DSL to look like, and I would be happy to change my implementation and create another pull request.

jszwedko:
Any update on this? #2586 looks compelling.

Otterpocket:
Was this resolved somewhere else? #2586 is still not merged

dfabreguette-ap:
I still have the same issue with latest version of mongoid & RailsAdmin :
rails_admin (2.2.1)
mongoid (7.3.3)

"unscoped" call removes all pagination filters in the Mongoid Criteria object.
Any idea how I can fix this as we use "default_scopes" on many rails models in our app, affecting default sorts.
As a result, sorting is simply not working in rails admin because of those default scopes.

[...]

PR #2428, opened 2015-10-09
Provide a config option for RailsAdmin to override default_scope

dburt:
Synopsis:

```ruby
RailsAdmin.config do |config|
  config.global_default_scope = :unscoped
  config.global_default_scope = ->{|model| model == User ? User.admins : model.all }
end
```

I've modelled my approach on this comment by @jxpx777 (https://github.com/sferik/rails_admin/issues/1348#issuecomment-39373394) but made it more flexible by allowing a block.

There are some things that still need to be resolved, which I need help with:

* Rubocop failures because I've added enough to the Config and Mongoid adapter modules that they've gone over 164 lines
* The best name for the config
* Documentation


mshibuya:
Alternative way already exists - please take a look at https://github.com/sferik/rails_admin/issues/1348#issuecomment-147045550.

dburt:
Tried it, it doesn't work. It only works for list views, and if you try to view or edit a record it breaks. Unless I'm missing something?

(See also this comment on Stack Overflow: this only works "if you only need to list the records": http://stackoverflow.com/a/28205484/1279840

mshibuya:
@dburt You're right, reopened #1348.

Anyway, I won't merge this in. I believe this kind of configuration should be achieved via instance option style DSL, not attr_accessor.

dburt:
Thanks for reopening. 
Can you please show me what you want the config file to look like? 
On 10/10/2015 8:24 PM, Mitsuhiro Shibuya <notifications@github.com> wrote:

@dburt You're right, reopened 
#1348.
Anyway, I won't merge this in. I believe this kind of configuration should be achieved via instance option style DSL, not attr_accessor.
—
Reply to this email directly or 
view it on GitHub.






This communication is confidential. If you are not the intended recipient of this email, any use, forwarding, printing or reproduction of it or any attachment, is prohibited. If you have received this communication in
 error, immediately contact us by return email. Whilst we employ an enterprise class anti-virus solution, we give no warranty express or implied that either this email or its attachments are free from viruses or other harmful code and we will accept no liability
 in respect of same. It is your responsibility to check for viruses or other harmful components before opening or using any attachments to this email. You should not copy, use, disclose or distribute this e-mail without the author's permission.

Please consider the environment and don't print this e-mail unless you really need to.

dalpo:
@mshibuya @dburt I think you could already manage it, using an authorization adapert like `cancancan`.

https://github.com/sferik/rails_admin/wiki/CanCanCan

For instance:

```ruby
class Ability
  include CanCan::Ability
  def initialize(user)
    if user.admin?
      can :manage, User, User.unscoped 
    else
      can :manage, User, User.another_scope
    end
  end
end
```

dburt:
Thanks @dalpo and I can see that that can work. But I don't think it's the best solution to use a separate authorization tool with 8 extra lines of configuration for each model for this simple, non-authorization-related use-case.

@mshibuya please show me what you want the DSL to look like, and I'll be happy to change my implementation.

dburt:
@mshibuya can this change be acceptable with a new DSL? If so, could you please show me what you'd like the DSL to look like? Then I could change the implementation. Otherwise, we could abandon and close this ticket.

MrOutis:
@dburt, I'm interested in this feature; maybe doing a [quick search for instance option style](https://github.com/sferik/rails_admin/search?utf8=%E2%9C%93&q=instance+option+style) can give you some clue to what @mshibuya refers to.

PR #2438, opened 2015-10-19
Fix multiselect javascript

ineu:
This fixes issue introduced in 0.7.0. Loop used indexes in `filtered` array instead of real value.


mshibuya:
Thanks :+1:

PR #2439, opened 2015-10-19
Don't assume english is always available

coorasse:


mshibuya:
It's giving an syntax error.

PR #2441, opened 2015-10-21
Eager load has_one and has_many in the list view so they can be searched.

bf4:
Eager load has_one and has_many in the list view so they can be searched.

### DIFF
--- a/app/controllers/rails_admin/main_controller.rb
+++ b/app/controllers/rails_admin/main_controller.rb
@@ -120,7 +120,7 @@ module RailsAdmin
     end
 
     def get_collection(model_config, scope, pagination)
-      associations = model_config.list.fields.select { |f| f.type == :belongs_to_association && !f.polymorphic? }.collect { |f| f.association.name }
+      associations = model_config.list.fields.select { |f| f.type.in?([:belongs_to_association, :has_one_association, :has_many_association]) && !f.polymorphic? }.collect { |f| f.association.name }
       options = {}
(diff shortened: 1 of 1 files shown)


mshibuya:
Always forcing eager load of `:has_many` association is not acceptable.
What if each of users has thousands of blog posts? Loading of user index page will be stuck.

app/controllers/rails_admin/main_controller.rb
```diff
@@ -120,7 +120,7 @@ def check_for_cancel
     end
 
     def get_collection(model_config, scope, pagination)
-      associations = model_config.list.fields.select { |f| f.type == :belongs_to_association && !f.polymorphic? }.collect { |f| f.association.name }
+      associations = model_config.list.fields.select { |f| f.type.in?([:belongs_to_association, :has_one_association, :has_many_association]) && !f.polymorphic? }.collect { |f| f.association.name }
```
bf4: is there another way we could do this?

bf4:
```ruby
def get_collection(model_config, scope, pagination)
  associations = _associations_for_collection(model_config)
  # etc
end
def _associations_for_collection(model_config)
  model_config.list.fields.select { |f| f.type == :belongs_to_association && !f.polymorphic? }.collect { |f| f.association.name }
end
```

so that we can customize it?

bf4: e.g. https://github.com/bf4/rails_admin/commit/53e53f49a41dece5f039b0b7bc98084468f76b60

PR #2446, opened 2015-10-28
Handle nullable foreign key in multiselect

jibidus:
`has_many` with not nullable foreign may raise exception when user try to remove elements from corresponding association.
The aim of this request is to prevent such behavior with disabling `remove` actions from form.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 ### Fixed
 - Add button for nested-many form in modal disappeared on click([#2372](https://github.com/sferik/rails_admin/issues/2372), [#2383](https://github.com/sferik/rails_admin/pull/2383))
+- Prevent users to remove association's element when foreign key is not nullable ([#2446](https://github.com/sferik/rails_admin/issues/2446))
 
 
 ## [0.7.0](https://github.com/sferik/rails_admin/tree/v0.7.0) - 2015-08-16
--- a/app/assets/javascripts/rails_admin/ra.filtering-multiselect.js
+++ b/app/assets/javascripts/rails_admin/ra.filtering-multiselect.js
@@ -17,6 +17,7 @@
         return { query: query };
       },
       sortable: false,
+      removable: true,
       regional: {
         up: "Up",
         down: "Down",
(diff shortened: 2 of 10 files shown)


mshibuya:
Thanks for your work! :beer:

arnvald:
@jibidus @mshibuya it seems that this PR breaks `has_many :through` relation :cry: 

I've got relation:

promotion has many user_promotions
promotion has many users through user_promotions

when I'm opening view to add new promotion, I can add users there, and what happens during this code execution:

```ruby
def foreign_key_nullable?
  return if foreign_key.nil? || type != :has_many
  klass.columns_hash[association.foreign_key].null
end
```

is that `foreign_key` is present (`:user_id`), relation type is `:has_many`, but `klass.columns_hash` does not contain `"user_id"` therefore it returns `nil` and the error is `undefined method "null" for nil:NilClass`

Any idea how to fix it?

jibidus:
@arnvald Let me have a look.

jibidus:
@arnvald I didn't manage to reproduce the issue :weary: .
When exactly the issue is raised? When you add user to new promotion ?
Do you have a RSpec test to be sure I can reproduce this issue?

arnvald:
@jibidus I created a simple application here: https://github.com/arnvald/rails_admin_polymorphic_bug

To reproduce:
1. `bundle install`
2. `bin/rake db:create db:migrate`
3. `bin/rails s`
4. go to `http://localhost:3000/admin/user/new`

I'll try to write a spec that'll cover this case, either today or tomorrow.

mshibuya:
@jibidus @arnvald Opened as a new issue: #2474

jibidus:
@arnvald @mshibuya Pull Request created: #2478.
Sorry for the issue and the delay :worried: .

arnvald:
@jibidus thanks for fixing the bug!

grillermo:
How can i configure this manually? i want some has_many fields without those buttons to remove the association. But adding 

        field :addresses do
          removable? do
            false
          end
        end

Does nothing.

jibidus:
This Pull Request was designed to prevent displaying the `Remove` button when foreign key is not nullable.
So, if make your foreign key not nullable, the `Remove` button will disappear.
If you don't want to make your foreign key not nullable, you cannot that (hide `Remove` button)

grillermo:
I made my own custom field for my use case, i don't want the foreign key nullable i wanted that field to not have that remove button on that particular model, so this is what i did:
https://gist.github.com/grillermo/9c0e5521a5e5b9361846128eb272c8c0
Thanks for the answer

PR #2448, opened 2015-10-31
Fix datetime parse logic

dalpo:
Consider actual Timezone and the Daylight Saving Time while parsing datetime input parameters.


dalpo:
Related to https://github.com/sferik/rails_admin/pull/2426

mshibuya:
Spec is failing, any ideas?

dalpo:
I will check soon!

dalpo:
@mshibuya Fixed and rebased from master but mongodb is still failing strangely... 
I will check mongodb in the next days.

dalpo:
@mshibuya now specs are green! :fireworks: 
But... we have a problem with the latest version of `tins` gem which actually has dropped the support to ruby 1.9.x https://github.com/flori/tins#changes

dalpo:
@mshibuya let me know what do you think about this PR!

Thanks

mshibuya:
Looks nice, thanks!

PR #2449, opened 2015-10-31
Fix regexp for converting rails date format to momentjs

prem-prakash:
Addresses https://github.com/sferik/rails_admin/issues/982


prem-prakash:
It does not work properly yet, I will refactor

PR #2453, opened 2015-11-03
Remove english dependency

coorasse:


lib/rails_admin/config/fields/types/datetime.rb
```diff
@@ -29,7 +29,7 @@ def parse_input(params)
           end
 
           register_instance_option :strftime_format do
-            fallback = ::I18n.t(date_format, scope: i18n_scope, locale: :en)
+            fallback = ::I18n.t(date_format, scope: i18n_scope)
```
mshibuya: This change does not make sense at all.
Since it's a fallback, you have to make sure that it works as a fallback in case of I18n retrieval failure.

mshibuya:
Please open another PR if you come up with another solution.

PR #2461, opened 2015-11-11
Added filtering for polymorphic association objects

arnvald:
Hi everyone! Today I've implemented this little piece for my application, I hope it will be useful enough to have it merged to original repo.

**Use case**

In my application I have a polymorphic relation. `Point` can belong either to `User` or a `Company`. The problem is that I have thousands of these records, so whenever I add a new `Point`, selecting proper `User` is a huge pain. Filtering polymorphic relations solves this issue.

**Demo**

polymorphic

**Implementation**

All the details are in the commit description: https://github.com/sferik/rails_admin/commit/1cc01e748db1cc49d3898f3a52b7da2d2cdae282

**Issues**

I've had a weird issue with CSS, that's why I had to add a `float_left` option to javascript. If anyone has better idea how to solve it, I'll be happy to try it out.

I've just noticed specs are failing on sqlite, I'll check it


arnvald:
Huh, I accidentally added a non-related commit to this branch (it fixes another issue I've had), and now tests are fixed. Not sure how to proceed further - should we keep this remporary fix?

@mshibuya any thoughts?

arnvald:
I'm still working on this one. To prevent website for freezing with large collections, for now I hardcoded `xhr` option to `true` and I preload only selected element. I'm planning to add advanced scoping, just like with simple, non-polymorphic select.

mshibuya:
Merging into #2886, thanks for contribution!

PR #2470, opened 2015-11-14
Fallback to default locale instead of :en

### LINKED ISSUE #2432: Assumptions that the English locale is available
There are a couple of assumptions here and there that the English locale is available.

This causes rails admin to fail (at least: 0.6.8, 0.7.0, master branch) in DateTime fields

Eg:

application.rb

``` rb
    config.i18n.default_locale = :it
    config.i18n.available_locales = %i(it)
```

``` rb
# Schema
create_table "players", force: :cascade do |t|
  t.string   "name"
  t.datetime "birthday"
end

# Player.rb
class Player < ActiveRecord::Base
  validates :name, presence: true
end

# Rails admin config
config.model Player do
  field :name
  field :birthday
end
```

Open rails admin, click on the model on the left sidebar, it will fail here:

https://github.com/sferik/rails_admin/blob/50d6e59bc8bf3515c84c78d874af66e6440fec5c/lib/rails_admin/config/fields/types/datetime.rb#L32

`I18n::InvalidLocale in RailsAdmin::Main#index`

tagliala:
Fix #2432

### DIFF
--- a/spec/rails_admin/config/fields/types/datetime_spec.rb
+++ b/spec/rails_admin/config/fields/types/datetime_spec.rb
@@ -25,6 +25,34 @@ describe RailsAdmin::Config::Fields::Types::Datetime do
 
       expect(field.formatted_value).to eq 'January 01, 2012 00:00'
     end
+
+    it 'handles non-English applications' do
+      begin
+        old_available_locales = I18n.available_locales
+        old_default_locale = I18n.default_locale
+        old_locale = I18n.locale
+
+        I18n.available_locales = [:fr]
+        I18n.default_locale = :fr
+        I18n.locale = :fr
+
+        RailsAdmin.config(FieldTest) do |_config|
+          field :datetime_field do
+            default_value DateTime.parse('01/01/2012')
+          end
+        end
+
+        field = RailsAdmin.config(FieldTest).fields.detect do |f|
+          f.name == :datetime_field
+        end.with(object: FieldTest.new)
+
+        expect(field.formatted_value).to eq 'dimanche 01 janvier 2012 00:00'
+      ensure
+        I18n.available_locales = old_available_locales
+        I18n.default_locale = old_default_locale
+        I18n.locale = old_locale
+      end
+    end
   end
 
   describe '#parse_input' do
--- a/lib/rails_admin/config/fields/types/datetime.rb
+++ b/lib/rails_admin/config/fields/types/datetime.rb
@@ -38,7 +38,7 @@ module RailsAdmin
           end
 
           register_instance_option :strftime_format do
-            fallback = ::I18n.t(date_format, scope: i18n_scope, locale: :en)
+            fallback = ::I18n.t(date_format, scope: i18n_scope, locale: ::I18n.default_locale)
             ::I18n.t(date_format, scope: i18n_scope, default: fallback).to_s
           end
 


spec/rails_admin/config/fields/types/datetime_spec.rb
```diff
@@ -25,6 +25,34 @@
...
+        end.with(object: FieldTest.new)
+
+        expect(field.formatted_value).to eq 'dimanche 01 janvier 2012 00:00'
+      ensure
+        I18n.available_locales = old_available_locales
+        I18n.default_locale = old_default_locale
+        I18n.locale = old_locale
+      end
+    end
   end
```
tagliala: Let me know if there is a better way of doing this.

I tried allow_any_instance_of but it only works when the test runs in isolation

dalpo:
@tagliala  did you consider the `lib/rails_admin/support/datetime.rb#L30` `delocalize` method? 

If you want remove totally the english locale dependency, I think you should also hardcode related english translations for:

- I18n.t('date.day_names', locale: :en)

- I18n.t('date.abbr_day_names', locale: :en)

- I18n.t('date.month_names', locale: :en)

- I18n.t('date.abbr_month_names', locale: :en)

- I18n.t('date.time.am', default: "am")

- I18n.t('date.time.pm', default: "pm")

cc: @mshibuya @sferik What do you think?

tagliala:
Didn't consider

tagliala:
Thanks for the suggestion, I will try another approach

tagliala:
I'm trying to understand how this works

### Step 1

I've searched for all `locale: :en` occurrences in the master and I've found 11 of them:

**rails_admin/lib/rails_admin/support/i18n.rb**
- https://github.com/sferik/rails_admin/blob/817e34bc6a653e3d4f0948dabeb63317be954ad8/lib/rails_admin/support/i18n.rb#L9
- https://github.com/sferik/rails_admin/blob/817e34bc6a653e3d4f0948dabeb63317be954ad8/lib/rails_admin/support/i18n.rb#L16
- https://github.com/sferik/rails_admin/blob/817e34bc6a653e3d4f0948dabeb63317be954ad8/lib/rails_admin/support/i18n.rb#L22
- https://github.com/sferik/rails_admin/blob/817e34bc6a653e3d4f0948dabeb63317be954ad8/lib/rails_admin/support/i18n.rb#L28
- https://github.com/sferik/rails_admin/blob/817e34bc6a653e3d4f0948dabeb63317be954ad8/lib/rails_admin/support/i18n.rb#L35

--

**rails_admin/app/views/rails_admin/main/history.html.haml**
- https://github.com/sferik/rails_admin/blob/817e34bc6a653e3d4f0948dabeb63317be954ad8/app/views/rails_admin/main/history.html.haml#L35

(question: is `unless object.created_at.nil?` different from `if object.created_at.present?`)

--

**rails_admin/lib/rails_admin/config/fields/types/datetime.rb**
- https://github.com/sferik/rails_admin/blob/022a4fe524b9c6fe99bda511b2f1a9dae76464d0/lib/rails_admin/config/fields/types/datetime.rb#L41

--

**rails_admin/lib/rails_admin/support/datetime.rb**
- https://github.com/sferik/rails_admin/blob/022a4fe524b9c6fe99bda511b2f1a9dae76464d0/lib/rails_admin/support/datetime.rb#L35
- https://github.com/sferik/rails_admin/blob/022a4fe524b9c6fe99bda511b2f1a9dae76464d0/lib/rails_admin/support/datetime.rb#L38
- https://github.com/sferik/rails_admin/blob/022a4fe524b9c6fe99bda511b2f1a9dae76464d0/lib/rails_admin/support/datetime.rb#L41
- https://github.com/sferik/rails_admin/blob/022a4fe524b9c6fe99bda511b2f1a9dae76464d0/lib/rails_admin/support/datetime.rb#L44

Issue #2474, opened 2015-11-17
#foreign_key_nullable? breaks with has_many through association


mshibuya:
https://github.com/sferik/rails_admin/pull/2446#issuecomment-156354184
https://github.com/sferik/rails_admin/pull/2446#issuecomment-157044663

PR #2478, opened 2015-11-18
Fix has_many through association

jibidus:
Fix `has_many` association with `through` issue introduced by #2446.
See #2446 for more details.

Issue fixed: #2474

### DIFF
--- a/lib/rails_admin/adapters/active_record/association.rb
+++ b/lib/rails_admin/adapters/active_record/association.rb
@@ -39,6 +39,7 @@ module RailsAdmin
 
         def foreign_key_nullable?
           return if foreign_key.nil? || type != :has_many
+          return true if association.through_reflection
           klass.columns_hash[foreign_key.to_s].null
         end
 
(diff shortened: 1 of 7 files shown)


lib/rails_admin/adapters/active_record/association.rb
```diff
@@ -39,6 +39,7 @@ def foreign_key
 
         def foreign_key_nullable?
           return if foreign_key.nil? || type != :has_many
+          return true if association.through_reflection
           klass.columns_hash[foreign_key.to_s].null
```
mshibuya: Are you sure that above fix is safe also for polymorphic association?
Isn't it safe to just nil-guard `null` call?
```ruby
klass.columns_hash[foreign_key.to_s].try(:null)
```

jibidus: This may return `nil` if key does not exist in `klass.columns_hash`. Perhaps not the return value we expect.
It should be better to write a test with a polymorphic association and see what happen (and fix potential issue), right ?
- [ ] Write test with polymorphic association

jibidus:
Thanks @mshibuya for your work.

PR #2479, opened 2015-11-19
XSS Fix: Use jQuery to build select options instead of +.

rogerbraun:
This fixes an XSS vulnerability with polymorphic associations.

Before, the label was used as directly as html instead of text.


sferik:
Thank you!

Issue #2481, opened 2015-11-23
Sass::SyntaxError in RailsAdmin::Main#dashboard with 0.8


mtparet:
I just upgraded an existing version from 0.7 to 0.8 and I got an error:
![selection_100](https://cloud.githubusercontent.com/assets/754770/11336383/2697eb66-91e5-11e5-9b1b-f556a9454cb7.jpg)

tnlin:
I have exactly the same problem!!!!!!!
debugging all night but got nothing
(This is my first time using RA, so depressive)

bensheldon:
It seems like #2391 (Replace jQueryUI datetimepicker with jQuery Bootstrap datetimepicker) added assets to `/vendor`, but they aren't being picked up in the SASS loadpaths. Also, it seems like the `/vendor` directory [was not added to the gemspec](https://github.com/sferik/rails_admin/blob/master/rails_admin.gemspec#L25) which may be why they aren't in the SASS loadpaths.

bensheldon:
FYI, the problem is the gemspec is not packaging the `vendor` directory. For a quick fix until this is fixed in the packaged version, you can change your gemfile declaration to:

`gem 'rails_admin', github: 'sferik/rails_admin'`

monkbroc:
I came to report this too. `vendor` is missing from the gemspec.

natebeaty:
+1. @bensheldon fix worked for me for now.

mshibuya:
Sorry about that :bow:
Just released the fix as 0.8.1.

monkbroc:
Arigatou Shibuya-san!

Thanks for your efforts in maintaining rails admin!

Hainminsoe:
i'm newbie in ror, please tell me . how can i fix it.

PR #2391, opened 2015-08-26
Replace jQueryUI datetimepicker with jQuery Bootstrap datetimepicker

dalpo:
- Refactor `Datetime`, `Date` and `Time` type inputs
- Replace jquery ui datepicker with https://eonasdan.github.io/bootstrap-datetimepicker/
- Refactor `Datetime`, `Date` and `Time` search filters

PR #2482, opened 2015-11-23
Add 'vendor' directory to gemspec

bensheldon:
It seems like #2391 (Replace jQueryUI datetimepicker with jQuery Bootstrap datetimepicker) added assets to `vendor`, but the `vendor` directory is not being packaged in the gem. This fixes that.

Fix for #2481.


dalpo:
My fault... I didn't realize that the vendor folder wasn't packaged in the gem...

:+1:

mshibuya:
Thanks!

PR #2491, opened 2015-11-27
Limit children for deletion notice

jibidus:
Limit number of future orphans displayed during deletion confirmation in order to avoid performance issue.

### DIFF
--- a/lib/rails_admin/abstract_model.rb
+++ b/lib/rails_admin/abstract_model.rb
@@ -91,7 +91,7 @@ module RailsAdmin
           end
         when :has_many
           children = object.send(association.name)
-          yield(association, children)
+          yield(association, Array.new(children))
         end
       end
     end
--- a/lib/rails_admin/abstract_model.rb
+++ b/lib/rails_admin/abstract_model.rb
@@ -87,12 +87,11 @@ module RailsAdmin
         case association.type
         when :has_one
           if child = object.send(association.name)
-            yield(association, child)
+            yield(association, [child])
           end
         when :has_many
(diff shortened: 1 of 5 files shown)


lib/rails_admin/abstract_model.rb
```diff
@@ -87,12 +87,11 @@ def each_associated_children(object)
...
           if child = object.send(association.name)
-            yield(association, child)
+            yield(association, [child])
           end
         when :has_many
-          object.send(association.name).each do |child| # rubocop:disable ShadowingOuterLocalVariable
-            yield(association, child)
-          end
+          children = object.send(association.name)
+          yield(association, Array.new(children))
```
alexandergitter: The problem with this is that it still fetches all children from the database.
I implemented a quick and dirty fix for this here: https://github.com/alexandergitter/rails_admin/commit/6a172da72c98bfd5ca7285b356478a16e7a7a064

jibidus: I don't think that children are fetched from database here.
`children` variable contains a `CollectionProxy` returned to `_delete_notice.html.haml` where `count` and `'first(limit)` is called. First n children are fetched here.
Am I wrong ?

alexandergitter: Mhh that's interesting, I didn't notice that.
In my tests Active Record did an unlimited SELECT though, not sure if there's something else that would trigger that... maybe it's because children gets wrapped in a new array? Would that eagerly fetch all children?

jibidus: Yes, wrapping in a a new array may fetch all children.

mshibuya:
Seems okay, thanks!

PR #2495, opened 2015-12-03
Fix parse BSON object value

dalpo:
This should fix the issue: https://github.com/sferik/rails_admin/issues/2489


dalpo:
This should fix https://github.com/sferik/rails_admin/issues/2489

mshibuya:
Thanks!

PR #2508, opened 2015-12-13
Improve encoding support in CSV converter

### LINKED ISSUE #2191: export csv give me error unknown encoding name - latin1
indipendent from the columns I choose I have always this error:

error unknown encoding name - latin1

tagliala:
Fix #2191

:warning: **Merge with extreme care** :warning:


bbenezech:
Thanks for this.

@tagliala Can I ask you why you need this?
It used to be back in the days some databases where not in utf-8, especially in asia, but I am wondering if it is needed at all today. What are you thoughts on this?

@mshibuya What should we do? utf8 all the things and stop this madness?

tagliala:
I run into the issue at #2191, there was a non-utf8 database whose tables could not be exported

Btw you could also drop the whole thing

bbenezech:
Ok, I read #2191, it seems a lot of people have this issue.
The code looks nice, let's merge this!

Thanks a lot @tagliala

PR #2509, opened 2015-12-16
json.rb now rescues JSON::ParserError

kevinryantao:
Fixing the bug raised in this issue:
crash on querying models with json columns
https://github.com/sferik/rails_admin/issues/2502


mshibuya:
Failing silently is bad, we have to tell the user that something is wrong with json.

Issue #1369, opened 2012-10-23
Blank multi-select


Jpunt:
I've got a basic has-many-through association:

project.rb:
```ruby
    has_many :project_assets, :dependent => :delete_all, :autosave => true, :include => :asset
    has_many :assets, :through => :project_assets
    attr_accessible :asset_ids
```

project_asset.rb:
```ruby
    belongs_to :asset
    belongs_to :project
```

asset.rb:
```ruby
    has_many :project_assets, :dependent => :delete_all
    has_many :projects, :through => :project_assets
```

I can add new assets through this successfully, but the left part of the multi-select remains blank, so I can't select assets that are already in the model. What could be the reason for this?

jwickard-taulia:
Spent 2 hours trying to figure out why my associations weren't showing up.  They were defined correctly, it was just this issue.

redterror:
Just wanted to pile on here - I'd really like to avoid explaining to my users that this isn't really a bug.  :)

daybreaker:
Throw one more on the pile of people who just spent way too long on this "issue" when all we need is a simple note somewhere around the search box. We have customer service reps who will be using this, and they probably wont remember the required functionality if they infrequently access that part of the edit form.

mshibuya:
How does the note should be read? I couldn't come up with a concise phrase...

Jpunt:
What actually needs to be done is some automatic paging inside the list and xhr-based searching. 

I'm not familiar with the RailsAdmin-codebase, so maybe someone who is can tell us if there already is some infrastructure in place to do something like this?

zenkalia:
Personally I would really like it if the multi select box was disabled and the body said something like "Too many objects, use search box above."

And if you type a query in and there are no results it would stay disabled and read "No objects found."

WillMasterJ:
It took me a while to find this post. I encountered the same issue.

mshibuya:
Added note as described above, thanks @zenkalia !

Jpunt:
Excellent news! 👍

zenkalia:
:heart:

SuperMasterBlasterLaser:
@mshibuya This problem still exists. How to help users to understand that he needs to type on input field?

mshibuya:
What version are you on?
Can you reproduce that in brand-new app?

SuperMasterBlasterLaser:
@mshibuya Version is 0.8.1

Many to one relationship where children are more than 100:

[...]

PR #2511, opened 2015-12-22
Added ability to change models number limit manually.

SuperMasterBlasterLaser:
According to #1369 RailAdmin set number of models that will be uploaded to multiselect forms. In other cases it will load nothing, forcing user to write something on input field to make search.

This feature makes users to be able to set number of limit manually:

     RailsAdmin.config do |config|
          config.association_select_models_number_limit = 500
     end

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -29,7 +29,7 @@ Metrics/MethodLength:
   Max: 29 # TODO: Lower to 15
 
 Metrics/ModuleLength:
-  Max: 164 # TODO: Lower to 100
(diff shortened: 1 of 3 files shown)


SuperMasterBlasterLaser:
Why all check failed? I didn't change core features of this gem.

SuperMasterBlasterLaser:
Tests says that my config file too long 166/164 for two lines :(

mshibuya:
Closing in favor of #2058, which comes with test code.

PR #2512, opened 2015-12-24
Allow `.parent` method to exist on models

cheerfulstoic:
If you define a `parent` method on your model then this line becomes a problem.  We can get the original method and bind it to our model like this.  Perhaps a bit klugey.  I'm open to other suggestions

### DIFF
--- a/lib/rails_admin/config/model.rb
+++ b/lib/rails_admin/config/model.rb
@@ -85,9 +85,8 @@ module RailsAdmin
         @navigation_label ||= begin
           # Use source method object in case it's overwritten in the model
           parent_module = Class.method(:parent).unbind.bind(abstract_model.model).call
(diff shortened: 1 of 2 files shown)


mshibuya:
Test needed!

cheerfulstoic:
Totally!  I just added a `.parent` class method to the `Player` ActiveRecord model in the specs.  I think this is causing things to fail still, so maybe it's not limited to just the place that I fixed.  I need to do other work at the moment, but I can look more later

mshibuya:
This became unnecessary because of https://github.com/sferik/rails_admin/commit/fd26cfb1f5ec1b89d3779982cd5df6176ff154d7#diff-bd5e679f57bbfa55594c30d60628e519R90.

PR #2525, opened 2016-01-11
Installation section improvments

GabrielLidenor:
The installation text should be more specific, we can't forget the beginners. :)


mshibuya:
Thanks!

PR #2539, opened 2016-01-26
Fix for bug: Uncaught Error: Syntax error

Aztre:
Fix for bug:
Uncaught Error: Syntax error, unrecognized expression: a[href=#unique-id- XXXXXXXXXXXXX ]
When it tries to remove a new nested entity.


bbenezech:
Thanks!

PR #2540, opened 2016-01-26
Fix momentjs translations for '%-d' format day of the month

ivanzotov:
Russian date long format has '%-d', so to_momentjs method translate "%-d %B %Y" into " MMMM YYYY"

https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/ru.yml#L43


mshibuya:
Thanks!

PR #2545, opened 2016-01-29
Added simple sorting by translated fields with globalize

FunkyloverOne:


mshibuya:
Please make sure that
- all the existing tests pass
- the changes you've just added are covered by new tests

FunkyloverOne:
I will add tests for my code, but that's pretty strange that lot of tests are falling even without my changes.
It seems like some tests are just old, because the code looks like it was improved. (The code results look smarter then expected results.)

bbenezech:
Sorry @FunkyloverOne but this sounds really too specific to Globalize.
Maybe you should instead provide a how-to with a config example?

PR #2547, opened 2016-01-31
Fix delocalize strftime_format for DateTime.strptime to support minus

ivanzotov:
DateTime.strprime can't parse formats that have '%-d' (https://github.com/svenfuchs/rails-i18n/search?utf8=%E2%9C%93&q=%25-d). So parse_string method returns nil and date field isn't saved correctly. I guess most users just overwrite date format to make rails_admin works properly, but this small fix let rails_admin works out of the box.


bbenezech:
Sounds good, thanks!

PR #2553, opened 2016-02-05
Not paginating when showing all history

minyoung:
When showing all history, we don't use paging (we're showing all history after all). But since Kaminari only adds extension methods when paging is used, so we end up in a situation where the paging methods are missing (which we assume are there). This happens when we're showing all history

Related issue:
https://github.com/sferik/rails_admin/issues/2503

### DIFF
--- a/app/views/rails_admin/main/index.html.haml
+++ b/app/views/rails_admin/main/index.html.haml
@@ -106,7 +106,7 @@
     - if @objects.respond_to?(:total_count)
       - total_count = @objects.total_count.to_i
       .row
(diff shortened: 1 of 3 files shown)


minyoung:
Note, there are no unit tests (yet). I haven't had the time to figure out how to add/run the tests yet

minyoung:
Now with tests (and removing unnecessary 'fix' to `app/views/rails_admin/main/index.html.haml`)

mshibuya:
Merged in, thanks!

Issue #2293, opened 2015-05-09
No route matches [POST] "/xxxx/3/edit" 


charly-palencia:
## Description
When i tried to edit some of my products or categories i got a POST RoutingError. I expect it sends the request to the  PUT action but it doesn't. what do i miss? PATH action probably? i'm not sure.
```
Started POST "/admin/product/3/edit" for ::1 at 2015-05-09 10:35:43 -0500

ActionController::RoutingError (No route matches [POST] "/admin/product/3/edit"):
  actionpack (4.2.0) lib/action_dispatch/middleware/debug_exceptions.rb:21:in `call'
  actionpack (4.2.0) lib/action_dispatch/middleware/show_exceptions.rb:30:in `call'
  railties (4.2.0) lib/rails/rack/logger.rb:38:in `call_app'
```

### Versions
**Rails** 4.2
**Rails_admin** 0.6.7

## router.rb
```ruby
  root :to => redirect('/admin')
  namespace :api do
    namespace :v1 do
      resources :categories
      resources :products
    end
  end

  mount RailsAdmin::Engine => '/admin', as: 'rails_admin'
```

### initializer/rails_admin.rb

```ruby

  config.actions do
    dashboard                     # mandatory
    index                         # mandatory
    new
    export
    bulk_delete
    show
    edit
    delete
    show_in_app
```

mshibuya:
Can you create minimal app to reproduce this problem?

charly-palencia:
@mshibuya  thanks for your replay. I fixed my issues while i was creating my new app :)
 
Apparently my problem was about a conflict with rails-api. I tried with [this](http://stackoverflow.com/questions/10923280/rails-admin-with-rails-api-conflict) but it doesn't work. So, i figured it out with this:

Gemfile
```
gem 'rails-api', require: false 
```

api_base_controller file

```
require 'rails-api'
```

Thanks you so much!

PR #2564, opened 2016-02-21
Explicity call the #t method on I18n

buren:
I'm building an API using `rails-api` and I had the same issue as described in #2293, so I added `require: false` to my Gemfile and explicitly called `require 'rails-api'` in my `ApplicationController`. 
When I then saved an record from with in rails admin I got an error `undefined method 't' for #<RailsAdmin::MainController:0x007f99a7bbc978>`. 
The fix is _very_ simple as you can see :)

thanks for an awesome gem! :star:


mshibuya:
Thanks!

Issue #1073, opened 2012-04-09
Ability to remove items from the navbar


dogweather:
In our installation, we believe that the navigation bar is very confusing for the end user. Our users are business managers, not techies:

![Screenshot](http://i.imgur.com/PYDqG.png)
- We want to remove the Gravatar. The default image is a very distracting bright blue that resembles an on/off button, inviting users to click it, without effect. We believe the gravatar only makes sense for certain user populations, and not ours.
- We want to remove the **Dashboard** link. We find it redundant, since the link on the far left fulfills the same role.
- Finally, in our app, we use the language, _sign in_ and _sign out_. Not _log in_ and _log out_. Devise also apparently uses this language by default. We need to standardize this, and this we need a way to remove or change the **Log out** button's text.

PR #2570, opened 2016-02-29
Making Gravatar optional in config

micdahl:
Added ability to disable Gravatar via config. Because including of the gravatar has moved from _secondary_navigation.html.haml to application_helper.rb disabling of the Gravatar icon in the Navbar has become very uncomfortable. So I have added config.show_gravatar which allows disabling the Gravatar via config.
#1073


bbenezech:
Thanks!

PR #2571, opened 2016-02-29
Update form_colorpicker.html.haml

nicovak:
The default css class of the field should be the bootstrap class "form-control"


mshibuya:
Thanks!

PR #2573, opened 2016-03-02
Update ra.filter-box.js

nicovak:
The label of the button should be included in the anchor, for the moment only the font awesome icon is clickable in the button.


mshibuya:
Thank you!

PR #2576, opened 2016-03-03
Fixes #2575 (pundit asking for all rows from models by monkey patching pundit

### LINKED ISSUE #2575: pundit extension inadvertently asks for all rows from models
`extensions/pundit/authorization_adapter.rb` contains this code:
```ruby
        def query(_action, abstract_model)
          @controller.policy_scope(abstract_model.model.all)
        rescue ::Pundit::NotDefinedError
          abstract_model.model.all
        end
```
but at least in `pundit 1.1.0`, the construction of `Pundit::NotDefinedError` calls `inspect` on the scope. This has the effect of retrieving every single row from the database table in question. We ran into this with a table that has over a million rows.

I have monkey-patched our application to get around this. I have also created a PR which has a spec that fails with the existing code and works with my patch. I'm submitting this issue before the PR so my commit can reference this issue.

ctm:
I'm not super happy with this PR, in that patching pundit seems ugly in general and adding a method whose name ends with `_or_nil` is gross in specific. However, the problem in the issue is real and this PR does resolve the problem.

I looked at pundit and didn't see a method that RailsAdmin could use to avoid the problem. Perhaps it makes sense to also submit a PR to pundit or at least open up an issue. Additionally, it's not obvious to me that having `ActiveRecord::Relation#inspect` retrieve all the rows of the relation is a particularly good idea, either.

So, I'm not at all wedded to this particular solution, but I am hoping that the issue is addressed so we can go back to using a stock RailsAdmin without monkey-patching.

### DIFF
--- a/config/initializers/pundit_patch.rb
+++ b/config/initializers/pundit_patch.rb
@@ -1,0 +1,20 @@
+# This file adds a method to the Pundit module that is logically
+# equivalent to Pundit#policy_scope, except when there is no
+# policy scope, nil is returned instead of raising a
+# Pundit::NotDefinedError.
+#
+# In addition to allowing us to avoid using exceptions as
+# flow-control, this bypasses any time consuming things that may occur
+# when the exception is created.
+#
+# In specific, this avoids sending inspect to an
+# ActiveRecord::Relation, which can be an exceedingly expensive
+# operation if that relation, for example, describes over a million
+# rows.
+if defined?(Pundit)
+  module Pundit
+    def policy_scope_or_nil(scope)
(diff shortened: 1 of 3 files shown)


ctm:
I see my code failed a `rubocop` check. I've fixed that and done a `push -f`.

ctm:
@mshibuya I see you have merged a few other PRs recently. Is there anything I can do to help get this one merged? Reading all rows into memory is prohibitively expensive.

mshibuya:
@ctm As you're saying, monkey-patching Pundit is ugly. If it is really needed, sending PR to Pundit is the right way to move.

mshibuya:
Please reopen on update.

PR #2586, opened 2016-03-15
Override default_scope globally with a config option

dburt:
This pull request is intended to resolve issue https://github.com/sferik/rails_admin/issues/1348 "add 'unscoped' mode to 'model config'" in a more general way.

Synopsis:

``` rb
RailsAdmin.config do |config|
  config.global_default_scope :unscoped
  config.global_default_scope { |model| model == User ? User.admins : model.all }
end
```

This is a revision of my earlier, rejected pull request to conform to @mshibuya's request for an "instance option style DSL, not attr_accessor": https://github.com/sferik/rails_admin/pull/2428#issuecomment-147068211


darrenterhune:
What happened with this? This is really awesome and would be nice if this merged. We have a project that could use this and we want to avoid monkey patching or forking 🍻

mshibuya:
I know the unscoped mode is of some use, but it should be a per-model configuration, rather than global one.

PR #2595, opened 2016-03-23
Remove Assyrian months from arabic translations

ramy-khater:
I just removed Assyrian months from arabic translations as in arabic world we do not use Assyrian calendar


mshibuya:
Could you submit PR to the [Moment.js](https://github.com/moment/moment) repository? We're just vendoring from it and we don't want to have our local change.
Once your change have been merged, then of course we can make vendored Moment.js up-to-date.

Fattouh92:
this is not solved yet, can you mention the pull request on moment.js repo

PR #2603, opened 2016-04-02
Adding dependency

weexpectedTHIS:
This is required for the assets to built properly with sass


mshibuya:
`font-awesome-sass-rails` is stale and we have switched to `font-awesome-rails` in fc314ef4caf87f5640a799088a713c8c2acaf7ec. This change makes no sense.

PR #2606, opened 2016-04-06
Add instance option to append a CSS class for rows

MrOutis:
As Rails Admin uses Bootstrap, I wanted to use the table's contextual classes in my lists.
This is an approach to implement that using a method declared in RailsAdmin::Config, it has access to the object being rendered into the row.
### Usage

``` ruby
RailsAdmin.config do |config|
config.model Event do
  list do
    row_css_class do
      bindings[:object].mode == 'panic' ? 'danger' : 'success'
    end
end
```

### DIFF
--- a/app/views/rails_admin/main/index.html.haml
+++ b/app/views/rails_admin/main/index.html.haml
@@ -90,7 +90,7 @@
           %th.last.shrink
       %tbody
         - @objects.each do |object|
-          %tr{class: "#{@abstract_model.param_key}_row"}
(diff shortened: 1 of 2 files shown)


mshibuya:
Test needed!

MrOutis:
indeed, I'll add them soon, @mshibuya; however, do you find it helpful?
I really don't know if there's a better/cleaner way to implement this.

mshibuya:
I think this is okay, it's simple and useful 😃

MrOutis:
@mshibuya, I will close this PR in favour of #2860.
It will be great if you can give me another review (after almost 1 year, hehe)

PR #2622, opened 2016-04-28
Duplicate filters on history nav

bartonmcguire:
There is a bug in which index page filters are duplicated when navigating back or forward in browser history. It's not a total show-stopper, but it's been bothering some of my users. And it is obviously incorrect/buggy behavior.

Repro steps:
1) Navigate to a RailsAdmin resource index.
2) Filter the index by some criteria (ex: 'Created At' => 'Is present')
3) Click "refresh" to apply the filter
4) Navigate somewhere within RailsAdmin (ex: click to "edit" a record in the index)
5) Navigate "back" via your browsers back button to the index.

Notice that there are now two "Created At" filters. Navigate forward and back again, now there are three... yikes!

This is because when navigating back the previously rendered filters are still on the page, but the inline script to `append` new filters gets run again.

The fix that I've come up with is to create an alternative function, `$.filters.init`, which is called inline instead. `init` simply checks whether any filters have been rendered already, and only builds filters for the array of options it is passed if they have not.

I've updated specs that test the view helper construction of the inline script. However, I found no tests anywhere that cover the behavior of `$.filters`, so I did not add tests for `$.filters.init`.

### DIFF
--- a/app/assets/javascripts/rails_admin/ra.filter-box.js
+++ b/app/assets/javascripts/rails_admin/ra.filter-box.js
@@ -3,6 +3,10 @@
   var filters;
 
   $.filters = filters = {
+    container: function() {
+      return $('#filters_box');
+    },
+
     append: function(options) {
       options = options || {};
       var field_label = options['label'];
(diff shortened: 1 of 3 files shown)


bartonmcguire:
This build is failing during gem installation - but clearly the diff shouldn't affect this. It looks like builds for other recent and unrelated pull requests are failing for the same reason - leading me to believe that my diff does not introduce the issue. Does `master` still successfully build? Has your CI settings/config changed since its last successful build?

cec:
I can confirm the reported issue, which many of our users are finding to be troublesome.
@mshibuya @bbenezech  I apologize for the direct mention as I know that u guys are very busy, but I'd like to know if u plan to address this and possibly backport it to earlier versions of the gem

mshibuya:
Fixed by #2998.

Issue #2623, opened 2016-04-28
Mongoid raise_not_found_error false cause undefined method for nil class


ppascualv:
Problem and workaround explained here: 

http://stackoverflow.com/questions/36911558/rails-admin-undefined-method-associations-for-nilnilclass/36911559#36911559

ericcf:
@pabliiitoo are you able to write a unit test for this issue?

pabliiitoo:
@ericcf I don't know how you test with diferent config files from 'wrapper' application.
If you can introduce me telling me which gems I need to meet or what part of code, I will try it.
Thank you.

dfabreguette-ap:
This is blocking for me in production - any idea when this can fixed ?

PR #2624, opened 2016-04-28
Fix issue #2623

pabliiitoo:


pabliiitoo:
Error given by ruby version and devise gem

lib/rails_admin/adapters/mongoid.rb
```diff
@@ -19,9 +22,12 @@ def new(params = {})
       end
 
       def get(id)
-        AbstractObject.new(model.find(id))
+        model_instance = model.find(id)
+        raise RaiseNotErrorFalse if model_instance.nil?
```
mshibuya: Introducing new error class is unnecessary, just returning here should be enough.

mshibuya:
Please add tests and resolve merge conflict, then reopen.

PR #2629, opened 2016-05-06
Add logout button when it has custom authorization

### LINKED ISSUE #2608: Logout button not showing.
I have set up my login using [manual authentication](https://github.com/sferik/rails_admin/wiki/Manually). It works great, but the logout button is not showing.

shinyaK14:
I've fixed #2608 issue.

**Before**:
screen shot 2016-05-06 at 5 48 54 pm

**After**:
screen shot 2016-05-06 at 6 07 39 pm

### DIFF
--- a/app/views/layouts/rails_admin/_secondary_navigation.html.haml
+++ b/app/views/layouts/rails_admin/_secondary_navigation.html.haml
@@ -8,3 +8,5 @@
       %li.edit_user_root_link= user_link
     - if logout_path.present?
       %li= link_to content_tag('span', t('admin.misc.log_out'), class: 'label label-danger'), logout_path, method: logout_method
+  - else
+    %li= link_to content_tag('span', t('admin.misc.log_out'), class: 'label label-danger'), main_app_root_path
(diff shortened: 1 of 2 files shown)


mohamedkamal-zanox:
+1

bbenezech:
Sorry, I don't get it. Why do you link to `main_app_root_path` when there is no `current_user`?
The link is already available there: https://github.com/shinyaK14/rails_admin/blob/cd0ce2c76c70ed106be629d251b3f5abfad5ac9c/app/views/layouts/rails_admin/_secondary_navigation.html.haml#L5

shinyaK14:
@bbenezech Thank you for your comment.
The reason is because when we set up [manual authentication](https://github.com/sferik/rails_admin/wiki/Manually), we cannot see logout button.
Yes, you are 100% correct. There is a link.
When user click the link and get back to root_path, we can consider as logout.

However the red logout button is more clear and user friendly.

mshibuya:
The button does not do 'logout', thus we don't state so.

PR #2633, opened 2016-05-15
Add support for grouping list by associated model

slbug:
There is no docs and tests. Also there is place for improvements. But currently I just want to ask if this option is useful. I use it to group my products by stores using same tabs as for scopes. Current usage:

``` ruby
class Product < ApplicationRecord
  belongs_to :store
end

class Store < ApplicationRecord
  has_many :producs
end

model.config Product do
  list do
    group_by { key: :store_id, model: Store }
  end
end
```

So if it's useful i will add docs/tests and do some refactoring. Suggestions are welcome.

P.S. Froala v2 update included :)


mshibuya:
Creating custom action will be better choice to accomplish this, not extending `index` action.

PR #2634, opened 2016-05-16
zh-CN support

iuhoay:


mshibuya:
Thanks for the PR, but we don't maintain translations in the repo.
Please take a look at(and update if necessary) [Translations](https://github.com/sferik/rails_admin/wiki/Translations) page in Wiki instead.

Issue #2329, opened 2015-06-12
Model named "Action" breaks RailsAdmin create


davidwparker:
I have a model named "Action".

When using RailsAdmin > "Actions" > create, it blows up with the error:

```
no implicit conversion of String into Hash
```

Stacktrace:
```
rails_admin (0.6.5) lib/rails_admin/config/actions/new.rb:26:in `merge'
rails_admin (0.6.5) lib/rails_admin/config/actions/new.rb:26:in `block (2 levels) in <class:New>'
(eval):8:in `instance_eval'
(eval):8:in `new'
actionpack (4.1.8) lib/action_controller/metal/implicit_render.rb:4:in `send_action'
actionpack (4.1.8) lib/abstract_controller/base.rb:189:in `process_action'
actionpack (4.1.8) lib/action_controller/metal/rendering.rb:10:in `process_action'
....
```
We're on RailsAdmin version 0.6.5.

(Link to 0.6.8, I don't think this changed since 0.6.5....)
https://github.com/sferik/rails_admin/blob/master/lib/rails_admin/config/actions/new.rb#L26

I believe that it has to do with the model name "Action", as I changed it to "Aaction" and it works. Is there a way to change this to work with the name "Action"? Thanks!

8398a7:
Model name(Action) is a duplicate of action in the param.
![2015-06-17 13 34 47](https://cloud.githubusercontent.com/assets/8043276/8199770/d9e86d16-14f5-11e5-8a5e-c1f4e3d94ecc.png)

This model name would be better to avoid.

onedanshow:
For posterity: Still an issue. Renamed my modal away from `Action` to resolve a `no implicit conversion of Symbol into Integer` issue when trying to visit the admin form to make a new record.

PR #2638, opened 2016-05-26
Ensure parameters is actually a Hash.

kakkoyun:
If you have a model named `Action` or a model that answer `action` to `#to_param` message, it causes a crash when trying to parse controller params as it explained #2329.

Proposed solution is an attempt to fix that issue.


kakkoyun:
Not a fully-fledged implementation, need more detailed consideration. Sorry.

PR #2642, opened 2016-05-28
added pundit policy authorized

Samy-Amar:
Pundit allows you to check if the authorize method was called for every page. Raises an error if it isn't. Rails Admin currently fails at it because it doesn't set the @_pundit_policy_authorized variable. This is just a quick fix. (taking inspiration from the rails_admin_pundit gem)


bf4:
What's a failure look like before this PR? Is a test possible?

Samy-Amar:
Before this PR, if you've told Pundit to *:verify_authorized* Pundit will raise an *AuthorizationNotPerformedError*, so the app crashes. I guess a test is possible, it would just check if the error appears or not

mshibuya:
Fixed by #2683.

Issue #2283, opened 2015-04-30
Problem with kaminari theme


benjaminwood:
I ran into an interesting problem today which is that I'm working on a project which uses both rails_admin & this gem: https://github.com/matenia/bootstrap-kaminari-views

The problem is that both have kaminari themes at this path: `views/kaminari/twitter-bootstrap`. So, when I'm using Kaminari in some non rails_admin view and specify the theme like this:

`= paginate @resource, remote: true, :theme => 'twitter-bootstrap'`

I end up with rails_admin kaminari theme instead of the other gem referenced above. The main issue with this is in regards to the `remote: true` option. When this option is passed, rails_admin adds a 'pjax' class and does not specify `data-remote=true` (standard rails/what I need).

My temp solution was to simply copy the bootstrap-kaminari-views theme into my app /views directory and title it something else.

Possible solution: change rails_admin kaminari theme name to something more specific like 'rails-admin-twitter-bootstrap'.

halilim:
Putting `gem 'bootstrap-kaminari-views'` *after* `gem 'rails_admin'` has an even more interesting effect: it removes styling from rails_admin pagination.

klyonrad:
So which gem is to blame here - who has to fix this?

benjaminwood:
I wouldn't say that either gem is to blame. It's simply a conflict between the two and technically either gem could fix it. Though, as I mentioned above I think it makes most sense that rails_admin changes the name of their theme, since the name of the other gem is `bootstrap-kaminari-views`. IE: I'm not sure what they would change it to.

If the maintainer(s) of rails_admin agree, perhaps one of us can submit a PR.

bbenezech:
@benjaminwood Sure, namespacing bootstrap-kaminari-views to ra-bootstrap-kaminari-views would make sense, no pb.

benjaminwood:
Cool, PR submitted ☝️

bbenezech:
merged!

PR #2651, opened 2016-06-08
Rename bootstrap-kaminari-views

benjaminwood:
See #2283 for more info.

tl;dr - [another gem](https://github.com/matenia/bootstrap-kaminari-views) uses the same kaminari theme name. This causes a conflict in apps that use both gems. This PR renames the kaminari views from twitter-bootstrap to ra-twitter-bootstrap.

Tests pass and I gave it a try in my browser too. Let me know if there is anything else that needs to be done here!


bbenezech:
Thanks a lot @benjaminwood !

benjaminwood:
No problem, thanks for your work and the merge! 🍻

PR #2659, opened 2016-06-23
bugfix: form_value use enum value not the key in active_record

dskyu:
when i use enum feature in my active record, the value of the select in form is '' because the value option of the select is enum's value not the key.
eg: there is a Product model with enum  status:  {spare: 1, on_sale: 2, off_sale: 3}
without this commit, 
image the status field is always null, because rails_admin form builder can't find the correct value for the select
with this commit,
image
it will find the correct value, so the select value will be filled correctly


luizpicolo:
This is fixed my error. Thanks.

PR #2664, opened 2016-06-27
Add viewport meta tag to mobile layout adjustment

luishmelo:
When using a smartphone, the layout do not adjust automatically, even with themes custom css because of the abscence of the viewport meta tag.


mshibuya:
Thanks!

PR #2665, opened 2016-06-27
Fix missing associated_model_config exception

Saicheg:
Hey!

CSV export did not work for me for `rails 5.0.rc2` release. Tracing that down i found that it was caused by type inconsistent ( comparing string vs symbol )

Please see attached screenshot.


mshibuya:
This PR was closed because I've deleted `rails-5` branch.
Please add tests and re-submit PR to the master if you still need this change.

PR #2670, opened 2016-07-11
Feature/allow multiple configuration blocks

semenovDL:
This pull request based by [Allow multiple config blocks for a single resource in RailsAdmin](https://github.com/sferik/rails_admin/pull/1781).
It implement the same functionality, but also support model expanding after initialization.
Specs included.


mshibuya:
Looks nice, thanks!

PR #2674, opened 2016-07-22
Ensure dashboard/statistics is being fully tested

jonduarte:
* Currently we are not checking if the statistics are really showing up.
  And since we are not adding the `index` action to the config block -
  which is mandatory - they actually will never appear, therefore the
  test will always pass, making the `statistics` flag useless.

  * Add an extra test to ensure we have both scenarios of showing or
  hiding statistics.


mshibuya:
Nice work, thanks!

Issue #1099, opened 2012-04-25
Is present and Is blank filter


kweejibow:
Apologies in advance I'm quite new to rails and rails_admin.

Seem like the 'Is Present' and 'Is blank' column filters fail for any boolean value

It also fails when checking against the integer value of a FK object. 

To better explain: I have a relationship between Client and Appointment, in my appointment list I include the field :client <-- where some automagick seems to display the clients name and create a link to that object. I can use all the other filters on the :client field just fine but if i try and check on Is Present or Is Blank the sql generated seems to try and check if client_id is not null and client_id != '' <-- which I believe is the issue.

Thanks!

nlsrchtr:
This is still relevant in `0.6.7`. Is there some work going into the Postgres specifics?

coorasse:
This error still exists. please reopen the issue.

jhilden:
also still having the issue with `0.6.8`

SuperMasterBlasterLaser:
This still exists on `0.8.1`

Hasan-a-ayar:
No fix yet? Same issue exists with 0.8.1

PR #2675, opened 2016-07-25
Issue #1099: Filtering on "Is present" and "Is blank" on boolean columns

HashNotAdam:
Filtering a boolean field on "Is present" or "Is blank" produces SQL that
includes checking against an empty string. Since a boolean field can only be
true, false, or null, this causes errors in some databases.

Remove the empty string search for boolean field types.

### DIFF
--- a/lib/rails_admin/adapters/active_record.rb
+++ b/lib/rails_admin/adapters/active_record.rb
@@ -152,22 +152,40 @@ module RailsAdmin
       end
 
       class StatementBuilder < RailsAdmin::AbstractModel::StatementBuilder
-      protected
+        protected
 
-        def unary_operators
-          {
-            '_blank' => ["(#{@column} IS NULL OR #{@column} = '')"],
-            '_present' => ["(#{@column} IS NOT NULL AND #{@column} != '')"],
-            '_null' => ["(#{@column} IS NULL)"],
-            '_not_null' => ["(#{@column} IS NOT NULL)"],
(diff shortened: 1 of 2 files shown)


HashNotAdam:
CI failing due to Rubocop issues addressed in [PR 2679](https://github.com/sferik/rails_admin/pull/2679)

mshibuya:
Test needed!

HashNotAdam:
Come on! The original code isn't tested... how about you test the existing functionality and I'll test the micro-change in this PR and keep using my fork in the mean-time? :P

mshibuya:
The original code is tested here,
https://github.com/sferik/rails_admin/blob/master/spec/rails_admin/adapters/active_record_spec.rb#L188-L198
just adding a few examples is enough. This is important to make sure the code you wrote won't break in the future.

HashNotAdam:
I've updated those tests but that is both nasty and terribly smelly. The fact that changing the code didn't cause a test to break tells you it wasn't properly tested.

There is no testing for the class itself, just the output after running through a module and two classes. If you are lucky enough for a future change to break the test, there will still be some analysis to work out where the issue exists. Even having 2 classes nested inside that module with module methods and classes mixed together is horrible and it makes it harder to find what you are looking for.

mshibuya:
I clearly state that horrible tests are better than no tests at all.
I know there's so many not-so-good things in the codebase, and since this is an open source software, it's everybody's job to improve them.

Thank you for your work, I really appreciate it 😃

Issue #2677, opened 2016-07-27
Autocomplete fields duplicated when using browser back/forward buttons


HashNotAdam:
When accessing the add or edit page for a model that uses the ui-autocomplete fields, using the back and forward buttons on the browser can cause the autocomplete fields to be duplicated.

I am experiencing this issue in Mac Chrome 51.0.2704.103

PR #2678, opened 2016-07-27
Issue #2677: Autocomplete fields duplicated when using browser back/forward buttons

HashNotAdam:
When accessing the add or edit page for a model that uses the ui-autocomplete fields, using the back and forward buttons on the browser can cause the autocomplete fields to be duplicated.

In _create, check if the field exists before creating it but still initialise autocomplete and set any necessary event handlers.

Also break out the Javascript into multiple, much smaller, functions so it is easier to follow. Since the Asset Pipeline can be used for magnification, we can afford to increase file size to increase readability. Who knows, the file may even get tested one day too.

### DIFF
--- a/app/assets/javascripts/rails_admin/ra.filtering-select.js
+++ b/app/assets/javascripts/rails_admin/ra.filtering-select.js
@@ -14,7 +14,9 @@
  *   jquery.ui.autocomplete.js
  */
 (function($) {
-  $.widget("ra.filteringSelect", {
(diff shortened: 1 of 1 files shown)


HashNotAdam:
CI failing due to Rubocop issues addressed in [PR 2679](https://github.com/sferik/rails_admin/pull/2679)

mshibuya:
Works fine, thanks!

PR #2679, opened 2016-07-27
Code review

HashNotAdam:
- Create better conformity to the style guide.
- Create compatibility with the Code Climate version of RuboCop to stop pull requests from being incorrectly labeled as failed.
- Add RuboCop ignores for constants that are being used a mutable items since adjusting this behaviour would require a lot of testing.
- DRY up some code seen in multiple places.

### DIFF
--- a/lib/rails_admin/config/fields/types/polymorphic_association.rb
+++ b/lib/rails_admin/config/fields/types/polymorphic_association.rb
@@ -16,7 +16,7 @@ module RailsAdmin
           # association checks that any of the child models are included in
           # configuration.
           register_instance_option :visible? do
-            associated_model_config.length > 0
+            !associated_model_config.empty?
           end
 
           register_instance_option :formatted_value do
--- a/lib/rails_admin/config/fields/types/polymorphic_association.rb
+++ b/lib/rails_admin/config/fields/types/polymorphic_association.rb
@@ -16,7 +16,7 @@ module RailsAdmin
           # association checks that any of the child models are included in
           # configuration.
           register_instance_option :visible? do
-            !associated_model_config.empty?
+            associated_model_config.any?
           end
 
           register_instance_option :formatted_value do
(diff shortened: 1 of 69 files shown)


lib/rails_admin/config/fields/types/polymorphic_association.rb
```diff
@@ -16,7 +16,7 @@ class PolymorphicAssociation < RailsAdmin::Config::Fields::Types::BelongsToAssoc
           # association checks that any of the child models are included in
           # configuration.
           register_instance_option :visible? do
-            associated_model_config.length > 0
+            !associated_model_config.empty?
```
tiii: !empty? == any?

spec/controllers/rails_admin/main_controller_spec.rb
```diff
@@ -113,7 +113,7 @@ def get(action, params)
 
   describe '#list_entries called from view' do
     before do
-      @teams = 21.times.collect { FactoryGirl.create :team }
+      @teams = Array.new(21) { FactoryGirl.create :team }
```
tiii: FactoryGirl.create_list(:team, 21)

tiii:
👍

lib/rails_admin/config/fields/types/polymorphic_association.rb
HashNotAdam: I'll be honest with you... I didn't like this but also didn't take the time to look into what associated_model_config returned and so played it safe :)
Looking at it now, it seems like `association.klass` should always return an array so it should be safe to use any?

I'll make this change now.

spec/controllers/rails_admin/main_controller_spec.rb
HashNotAdam: Good call!

mshibuya:
Excellent work, thanks a lot!

HashNotAdam:
Wonderful, thank you, I'll rebase my other PRs now.

PR #2680, opened 2016-07-28
Support for ActiveRecord:Enum using string columns

tiii:
ActiveRecord::Enum supports storing values in string database columns. This behaviour is not yet described in the rails documentation, but recently a test case for it was added. (https://github.com/rails/rails/commit/67c1719012506c3387df067961252b5df50a97ce)

I did not add extra tests for it, since the active_record_enum Type has not been tested previously. I did however add ActiveRecord::Enum attributes (with and without string columns) to the test models, so the behaviour is at least tested implicitly.


mshibuya:
Amazing, thanks!

PR #2693, opened 2016-08-17
Added option list to disable link "Add filter"

luizpicolo:
I added an option to disable the link "Add filter" if the developer does not wish the option is visible desired model.

Without the option, or to set it to true, the option will appear normally.
```ruby
config.model 'About' do
  list do
    field :id
    field :title
  end
end
```
or 

```ruby
config.model 'About' do
  add_filter true
  list do
    field :id
    field :title
  end
end
```
screenshot from 2016-08-17 14-40-41

But, if set false, will not be show in the model list .

```ruby
config.model 'About' do
  add_filter false
  list do
    field :id
    field :title
  end
end
```

screenshot from 2016-08-17 14-40-412

### DIFF
--- a/app/helpers/rails_admin/main_helper.rb
+++ b/app/helpers/rails_admin/main_helper.rb
@@ -37,7 +37,7 @@ module RailsAdmin
     end
 
     def filterable_fields
-      @filterable_fields ||= @model_config.list.fields.select(&:filterable?)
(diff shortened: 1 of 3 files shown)


mshibuya:
This can be achieved by marking all fields as `filterable false`.

luizpicolo:
The problem is when the list of attributes and models becomes very large . I know this methods , but that the PR would be a facilitator for all attributes and not a filter for them.

PR #2701, opened 2016-08-25
pass parameters to new forms when model is namespaced

loicginoux:
When model is namespaced, the prepopulation of the new form via params didn't work.
That should fix it.

### DIFF
--- a/spec/integration/basic/new/rails_admin_namespaced_model_new_spec.rb
+++ b/spec/integration/basic/new/rails_admin_namespaced_model_new_spec.rb
@@ -33,6 +33,5 @@ describe 'RailsAdmin Namespaced Model New', type: :request do
       visit new_path(model_name: 'cms~basic_page', cms_basic_page: {title: 'Hello'})
       expect(page).to have_css('input[value=Hello]')
     end
-
   end
 end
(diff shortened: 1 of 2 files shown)


mshibuya:
Why is this fix needed? Isn't it enought to pass `'cms~basic_page' => {title: 'Hello'}`?

loicginoux:
No it didn't work, I committed a test for this case.

lazebny:
@mshibuya i see the same issue.

lazebny:
@loicginoux could you rebase these commits in one ? (might this is blocker for merge)

spec/integration/basic/new/rails_admin_namespaced_model_new_spec.rb
```diff
@@ -27,4 +27,11 @@
       is_expected.to have_selector("textarea#cms_basic_page_content[name='cms_basic_page[content]']")
     end
   end
+
+  describe 'GET /admin/cms_basic_page/new with parameters for pre-population' do
+    it 'populates form field when corresponding parameters are passed in' do
+      visit new_path(model_name: 'cms~basic_page', cms_basic_page: {title: 'Hello'})
```
lazebny: You are visit page twice, once in before block, second time in it block. Could you fix this?

loicginoux: which before block do you mean ? the one in line 20 is not called in this test.

lazebny: Oh sorry, you are right.

Eversilence:
I am having the same issues with forms.. Would be great if proposed PR would be merged

mshibuya:
Merged in, thanks!

PR #2703, opened 2016-08-27
Anonymous class compatibility

patleb:
When using `active_type` gem, `ActiveType::Record[AnyModel]` parent class is anonymous, so its name is nil.


mshibuya:
We need test code to make sure this does not break in the future, passing current spec is not enough.

PR #2712, opened 2016-09-15
Must use page_method_name for all Kaminari pagination

carsonreinke:
Fixes https://github.com/sferik/rails_admin/issues/2711


mshibuya:
Spec is failing, can you make it pass?

carsonreinke:
Yeah, Kaminari defines the pagination method statically like this:

```
    # items at the specified "page"
    class_eval <<-RUBY, __FILE__, __LINE__ + 1
      def #{Kaminari.config.page_method_name}(num = 1)
        offset(limit_value * ((num = num.to_i - 1) < 0 ? 0 : num))
      end
    RUBY
```

So really, tests were wrong before.  Their classes would have to be forcefully reloaded to change the pagination method on the fly.

carsonreinke:
Without having a fork of some sort, it is not really possible to dynamically change the Kaminari page method.  The ActiveRecord portion is even more complicated to try and reload.

I see 3 tests that are incorrectly using this approach.

Any advice, thoughts, wisdom to get something setup would be appreciated.

mshibuya:
Added a spec in https://github.com/sferik/rails_admin/commit/73ce3251896e668c82d553c5f612d5abace95b24, thanks.

PR #2715, opened 2016-09-16
rewrite selected and queryFilter fn in ra.filtering-multiselect.js

xoxoxo:
Had some performance issues with multiselect that have loads of objects. In our case 10k we could load ajax but decided to speed up this plugin. Its up to 5-10x faster.


mshibuya:
Fantastic, thanks!

PR #2720, opened 2016-09-25
Do not assume that every model supports Dragonfly

skalee:
It is not guaranteed that every model extends `Dragonfly::Model`. Calling `::dragonfly_attachment_classes` on some models would raise `NoMethodError`.


mshibuya:
Spec needed!

mshibuya:
Added a spec in https://github.com/sferik/rails_admin/commit/93510ffec8b30f012cc053642d1fe7b20bbbcbc2, thanks.

PR #2737, opened 2016-10-12
fixes 'Cannot set property '_renderItem' of undefined' when adding a nested resource

### LINKED ISSUE #2722: Accepts_nested_attributes_for not working (in v 1.0.0)
Firstly, AWESOME JOB with version 1.0.0! :) :) :)

The example:
1. accepts_nested_attributes_for :comments
2. Comment model has belongs_to :post  (so should create a ui-autocomplete-input)..but gives the error: 
"Uncaught TypeError: Cannot set property '_renderItem' of undefined"

Please see attached pic. 

I tried with different projects and models and nothing worked. 

Rails admin 1.0.0
Rails 5.0.1

screen shot 2559-09-26 at 9 25 00 pm

Thanks again! :)

tmesquita:
Fixes #2722

### DIFF
--- a/app/assets/javascripts/rails_admin/ra.filtering-select.js
+++ b/app/assets/javascripts/rails_admin/ra.filtering-select.js
@@ -33,22 +33,21 @@
     select: null,
 
     _create: function() {
-      var filtering_select;
+      var filtering_select = this.element.siblings(
+        '[data-input-for="' + this.element.attr('id') + '"]'
+      );;
 
       // When using the browser back and forward buttons, it is possible that
       // the autocomplete field will be cached which causes duplicate fields
       // to be generated.
-      if (this.element.is(':visible')) {
+      if (filtering_select.size() > 0) {
+        this.input = filtering_select.children('input');
+        this.button = filtering_select.children('.input-group-btn');
+      } else {
         this.element.hide();
         filtering_select = this._inputGroup(this.element.attr('id'));
         this.input = this._inputField();
         this.button = this._buttonField();
-      } else {
-        filtering_select = this.element.siblings(
-          '[data-input-for="' + this.element.attr('id') + '"]'
-        );
-        this.input = filtering_select.children('input');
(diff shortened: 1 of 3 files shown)


mshibuya:
Spec is failing, can you take a look?

tmesquita:
I'm looking into, it looks like it fails when using Capybara >=2.10.0. I haven't used Capybara much so it's taking me a little while trying to figure out what in that update is causing the test to no longer pass

tmesquita:
I think the test that is failing is a bum test

From what I can tell, the block for `have_selector` was not getting run prior to Capybara v2.1.0 which was released about a week ago (2016-10-05) since that functionality wasn't implemented until that version (https://github.com/jnicklas/capybara/commit/b5655e9b9bde69aa9fba19902334b76c99f62148#diff-5b48908a3d2bd3374aaef7b6394f1912R39)

```
# https://github.com/sferik/rails_admin/blob/master/spec/integration/basic/edit/rails_admin_basic_edit_spec.rb#L76
describe 'edit with has-and-belongs-to-many association' do
    before do
      @teams = FactoryGirl.create_list(:team, 3)
      @fan = FactoryGirl.create :fan, teams: [@teams[0]]
      visit edit_path(model_name: 'fan', id: @fan.id)
    end

    it 'shows associated objects' do      
      is_expected.to have_selector '#fan_team_ids' do |select|         
        # This block doesn't seem to run prior to Capybara v2.1.0
        expect(select[0]).to have_selector 'option[selected="selected"]'
        expect(select[1]).not_to have_selector 'option[selected="selected"]'
        expect(select[2]).not_to have_selector 'option[selected="selected"]'
      end
    end
end
```

benhutton:
@tmesquita thank you! :+1:

kubudubu:
When can we expect integration with master? ;)

mshibuya:
Awesome, thanks!

PR #2738, opened 2016-10-13
Fixing test relations

tmesquita:
When a model has a "test" relation, like the below example, navigating to the model's show page throws an `ArgumentError wrong number of arguments (0 for 2..3)` error. 
```
class Case < ActiveRecord::Base
  belongs_to :test, inverse_of: :cases
end
```
After digging around from what I can tell it's from the use of the `send` method in `lib/rails_admin/config/fields/association.rb`. 
The ActiveRecord::Base object appears to have a private `test` method that is being invoked before the public `test` method that retrieves the association. 

I've updated it to use `public_send` instead. I can't think of any reason where a private method would be used when retrieving the value for the association.

### DIFF
--- a/lib/rails_admin/config/fields/association.rb
+++ b/lib/rails_admin/config/fields/association.rb
@@ -98,7 +98,7 @@ module RailsAdmin
 
         # Reader for the association's value unformatted
         def value
-          bindings[:object].send(association.name)
+          bindings[:object].public_send(association.name)
         end
 
         # has many?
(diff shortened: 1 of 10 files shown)


tmesquita:
looking into failing specs

tmesquita:
closing since test seems to be a reserved word in Rails

tmesquita:
I'm actually flip-flopping, while my use case in which this bug appeared might not be a good one, it still might be beneficial to use `#public_send` instead of `#send`. Maybe I should remove the added test case but keep the PR. Thoughts @mshibuya?

PR #2741, opened 2016-10-13
csv export options need to be HashWithIndifferentAccess

loicginoux:
a correction for this bug.
https://github.com/sferik/rails_admin/issues/2740

### DIFF
--- a/lib/rails_admin/support/csv_converter.rb
+++ b/lib/rails_admin/support/csv_converter.rb
@@ -32,6 +32,7 @@ module RailsAdmin
     end
 
     def to_csv(options = {})
+      options = HashWithIndifferentAccess.new(options)
       encoding_to = Encoding.find(options[:encoding_to]) if options[:encoding_to].present?
 
       csv_string = generate_csv_string(options)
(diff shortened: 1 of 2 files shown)


lib/rails_admin/support/csv_converter.rb
```diff
@@ -32,6 +32,7 @@ def initialize(objects = [], schema = {})
     end
 
     def to_csv(options = {})
+      options = HashWithIndifferentAccess.new(options)
```
mshibuya: Calling [`symbolize_keys!`](http://api.rubyonrails.org/classes/Hash.html#method-i-symbolize_keys-21) would be a bit simpler, I guess.

mshibuya:
Test code needed.

lib/rails_admin/support/csv_converter.rb
loicginoux: I guess this is easier but I prefered to use the safest option. First, not knowing the rest of the code and the consequences, I prefered to be able to use string and symbol as keys, to be safe. Also, you suggestion would actually be to use `.deep_symbolize_keys!` 
At least using `HashWithIndifferentAccess`, I am sure I am not introducing any regression.

mshibuya:
Can be merged when rubocop offenses are resolved.
https://travis-ci.org/sferik/rails_admin/jobs/174276072#L425-L430

loicginoux:
done.

mshibuya:
Thank you!

PR #2747, opened 2016-10-23
More languages

rubenandre:
We could add more languages for this rails admin, I leave the Portuguese translation.


mshibuya:
We maintain all locales in Wiki, add it there if you like.

PR #2751, opened 2016-10-31
Association filters

impurist:
Proposed expansion of the field DSL to allow better association filtering.
Expanded and refactored filter box JavaScript. 
  DSL Example:
 See [rails_admin_basic_list_spec](https://github.com/impurist/rails_admin/blob/association-filters/spec/integration/basic/list/rails_admin_basic_list_spec.rb) for more examples.
```
config.model Team do
        field :name
        field :fans, :has_and_belongs_to_many_association do
        searchable [:id]
        filterable true

        filter_by do
           Fan.all.map { |c| [c.name, c.id] }
        end

        multi_select do
            true
        end

        check_boxes do
           true
        end
    end

    list do
       filters [:fans]
    end
end
```
I value any comments, suggestions, criticisms or indeed help to get these proposed changes right. and accepted.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ tmp/**/*
 .idea/*.xml
 .sass-cache
 nbproject
+/.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ tmp/**/*
 .sass-cache
 nbproject
 /.DS_Store
+*/**/.DS_Store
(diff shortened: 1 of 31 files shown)


impurist:
The Contribution guide should point out the Appraisals builds and ESLint.

mshibuya:
Sorry for long quietness.
Your refactoring seems to be pretty opinionated, could you separate addition of association filtering and javascript refactoring?

mshibuya:
Please let me know when ready.

PR #2752, opened 2016-11-02
Config option to change default csv encoding

omjokine:
Option to configure CSV export encoding. 

Currently default is Rails default, which is usually UTF-8. That's usually fine, but Windows Excel doesn't support it and list of available encodings is very long so it's not too easy to find correct Windows-xyz encoding when you are doing CSV export on Windows machines. So I added config option (csv_default_encoding) to change default encoding used in export menu.


mshibuya:
This should be `export` action's option, not global one.

Issue #2756, opened 2016-11-04
NoMethodError: undefined method `parse_value' for nil:NilClass when using custom filter parameters


leegeng:
Ruby: 2.3.1
Rails: 4.2.7
OS: MacOS Sierra 10.12.1
Gem: 1.0.0

I'm trying to make a custom filter on a model's list. 
In this case, I put some parameters on ```f``` hash. (some of them are defined as a field in config, and the others are not because their origin database is different. I wanted to remake query on index_extension.)

I got an error like this because field parameter is nil. (https://github.com/sferik/rails_admin/blob/master/lib/rails_admin/adapters/active_record.rb#L140)

```
NoMethodError: undefined method `parse_value' for nil:NilClass
  from rails_admin/abstract_model.rb:114:in `parse_field_value'
  from rails_admin/adapters/active_record.rb:140:in `block (2 levels) in filter_scope'
  from action_controller/metal/strong_parameters.rb:185:in `each_pair'
  from action_controller/metal/strong_parameters.rb:185:in `each_pair'
  from rails_admin/adapters/active_record.rb:137:in `block in filter_scope'
  from action_controller/metal/strong_parameters.rb:185:in `each_pair'
  from action_controller/metal/strong_parameters.rb:185:in `each_pair'
  from rails_admin/adapters/active_record.rb:136:in `filter_scope'
  from rails_admin/adapters/active_record.rb:34:in `all'
  from rails_admin/main_controller.rb:131:in `get_collection'
  from rails_admin/main_controller.rb:37:in `list_entries'
  from lib/rails_admin/config/actions/index_extension.rb:12:in `block in controller'
  from rails_admin/main_controller.rb:22:in `instance_eval'
  from rails_admin/main_controller.rb:22:in `index'
  from action_controller/metal/implicit_render.rb:4:in `send_action'
  from abstract_controller/base.rb:198:in `process_action'
  from action_controller/metal/rendering.rb:10:in `process_action'
  from abstract_controller/callbacks.rb:20:in `block in process_action'
  from active_support/callbacks.rb:117:in `call'
  from active_support/callbacks.rb:555:in `block (2 levels) in compile'
  from active_support/callbacks.rb:505:in `call'
  from active_support/callbacks.rb:92:in `__run_callbacks__'
  from active_support/callbacks.rb:778:in `_run_process_action_callbacks'
  from active_support/callbacks.rb:81:in `run_callbacks'
[...]

PR #2757, opened 2016-11-04
#2756 fixed nil error for custom filter parameter

leegeng:
detail explanation is on issue #2756


mshibuya:
Test needed!

mshibuya:
Please reopen on update.

PR #2766, opened 2016-11-14
Search query: Added ILIKE support for PostGIS adapter, multibyte downcase for other adapters; support for UUID

tinbka:
Since 02.02.2017, this PR also addresses complaints from https://github.com/sferik/rails_admin/issues/1681 (UUID search)


mshibuya:
Looks good, thanks!

PR #2771, opened 2016-11-15
Load models from eager_load, not autoload_paths

### LINKED ISSUE #2770: Rails Admin loads every single file in autoload paths!
Every single file in `autoload_paths` is eagerly loaded by Rails admin. Below, I'll explain how this happens, why this shouldn't happen, and will propose a solution.

I came across this when trying to figure a bizarre issue in Thredded + rails_admin compatibility: https://github.com/thredded/thredded/issues/280.

## How does this happen?

Here is what RailsAdmin tries to load, [lib/rails_admin/config.rb#L322](https://github.com/sferik/rails_admin/blob/ffa30e0563e86a91bf88a312911b77d1bfffc2dc/lib/rails_admin/config.rb#L322):

```ruby
def viable_models
  included_models.collect(&:to_s).presence || begin
    @@system_models ||= # memoization for tests
        ([Rails.application] + Rails::Engine.subclasses.collect(&:instance)).flat_map do |app|
          (app.paths['app/models'].to_a + app.config.autoload_paths).collect do |load_path|
            Dir.glob(app.root.join(load_path)).collect do |load_dir|
              Dir.glob(load_dir + '/**/*.rb').collect do |filename|
                # app/models/module/class.rb => module/class.rb => module/class => Module::Class
                lchomp(filename, "#{app.root.join(load_dir)}/").chomp('.rb').camelize
              end
            end
          end
        end.flatten.reject { |m| m.starts_with?('Concerns::') } # rubocop:disable MultilineBlockChain
  end
end
```

It loads these by calling constantize in [lib/rails_admin/abstract_model.rb#L13-L25](https://github.com/sferik/rails_admin/blob/9e1c66ee036c640ac1b65aab35947c2ebd2671ef/lib/rails_admin/abstract_model.rb#L13-L25):

```ruby
def all(adapter = nil)
  @@all ||= Config.models_pool.collect { |m| new(m) }.compact
  adapter ? @@all.select { |m| m.adapter == adapter } : @@all
end

alias_method :old_new, :new
def new(m)
  m = m.constantize unless m.is_a?(Class)
  (am = old_new(m)).model && am.adapter ? am : nil
rescue LoadError, NameError
  puts "[RailsAdmin] Could not load model #{m}, assuming model is non existing. (#{$ERROR_INFO})" unless Rails.env.test?
  nil
end
```

### Why this shouldn't happen?

If something is in `autoload_paths`, it doesn't mean it can or should *always* be loaded. Rails explicitly makes this distinction with `eager_load`.

E.g. if I'm developing a Rails engine, I might want to add `lib` to autoload paths (to ease the development of the engine), but that doesn't mean I want to eagerly load ruby files in `lib/generators/templates`!

### Proposed solution

There are a few options here:

1. Remove `app.config.autoload_paths` from the files RailsAdmin tries to autoload. Only load things from `app/models` directories.
2. Replace `app.config.autoload_paths` with `app.paths.eager_load`.
3. Stop messing with the loading system! Just include a module into the base class!

3 is ideal but is also the most effort and I don't have the time. 1 breaks backwards compatibility.
So I'll send a PR for 2 for now.
Perhaps one day somebody can do 3.

@mshibuya @bbenezech @sferik

glebm:
If something is in `autoload_paths`, it doesn't mean it can or should *always* be loaded. Rails explicitly makes this distinction with `eager_load`.

E.g. if I'm developing a Rails engine, I might want to add `lib` to autoload paths (to ease the development of the engine), but that doesn't mean I want to eagerly load ruby files in `lib/generators/templates`!

Resolves #2770

### DIFF
--- a/lib/rails_admin/config.rb
+++ b/lib/rails_admin/config.rb
@@ -323,7 +323,7 @@ module RailsAdmin
         included_models.collect(&:to_s).presence || begin
           @@system_models ||= # memoization for tests
             ([Rails.application] + Rails::Engine.subclasses.collect(&:instance)).flat_map do |app|
-              (app.paths['app/models'].to_a + app.config.autoload_paths).collect do |load_path|
+              (app.paths['app/models'].to_a + app.paths.eager_load).collect do |load_path|
                 Dir.glob(app.root.join(load_path)).collect do |load_dir|
                   Dir.glob(load_dir + '/**/*.rb').collect do |filename|
                     # app/models/module/class.rb => module/class.rb => module/class => Module::Class
--- a/spec/dummy_app/config/application.rb
+++ b/spec/dummy_app/config/application.rb
@@ -20,7 +20,7 @@ module DummyApp
     # Application configuration should go into files in config/initializers
     # -- all .rb files in that directory are automatically loaded.
     config.eager_load_paths.reject! { |p| p =~ %r{/app/(\w+)$} && !%w(controllers helpers views).push(CI_ORM).include?(Regexp.last_match[1]) }
-    config.autoload_paths += %W(#{config.root}/app/#{CI_ORM} #{config.root}/app/#{CI_ORM}/concerns)
+    config.autoload_paths += %W(#{config.root}/app/#{CI_ORM} #{config.root}/app/#{CI_ORM}/concerns #{config.root}/lib)
     config.i18n.load_path += Dir[Rails.root.join('app', 'locales', '*.{rb,yml}').to_s]
     config.active_record.raise_in_transactional_callbacks = true if Rails::VERSION::MAJOR == 4 && Rails::VERSION::MINOR == 2 && CI_ORM == :active_record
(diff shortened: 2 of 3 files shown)


impurist:
This makes sense to me. I currently have to explicitly exclude classes in lib. Annoying!
 Your idea is valid.However!
eager_load is about table names in relational queries with ActiveRecord and not about loading model classes with Ruby. These are 2 very different little beasts.

glebm:
@impurist I think you're confusing ActiveRecord eager_load with config.eager_load_paths. They are unrelated.

glebm:
@mshibuya どう思いますか？

impurist:
So eager_load_paths is poorly named

mshibuya:
Agreed, but it would be better to have a spec that will fail without this change.

glebm:
@mshibuya Done

mshibuya:
Nice, can be merged if [the rubocop offense](https://travis-ci.org/sferik/rails_admin/jobs/212459733#L682) is fixed.

glebm:
@mshibuya Sorry, didn't notice it. Now fixed!

mshibuya:
Thanks for the fix!

Issue #2775, opened 2016-11-19
Status code 404 when record fails to be destroyed from the index page


chamun:
When I try to destroy a record from the index page and the record fails to be destroyed, I am redirected back to the index page with status code 404.

I expect to be redirected to the index page with status code 200, as it happens when I try to destroy the same record from the show page.

The issue is related on how the destroy and the index action interact with the `flash[:error]`.

On [/lib/rails_admin/config/actions/delete.rb:40](https://github.com/sferik/rails_admin/blob/9e1c66ee036c640ac1b65aab35947c2ebd2671ef/lib/rails_admin/config/actions/delete.rb#L40) the `flash[:error]` is set with the error message after the object destruction fails.

On [/lib/rails_admin/config/actions/index.rb:44](https://github.com/sferik/rails_admin/blob/9e1c66ee036c640ac1b65aab35947c2ebd2671ef/lib/rails_admin/config/actions/index.rb#L44) the status code `:not_found` is returned if `flash[:error]` is present, which is not the right decision in this situation.

abhi-sm:
Please close this issue.

PR #2776, opened 2016-11-19
Return status code 200 when destroying a record from the index page fails

chamun:
This PR closes issue #2775.

It changes how the index and the dashboard actions decide which status code to return. Instead of relying on `flash[:error]` to render status code 404, the new instance variable `@status_code` is used.


mshibuya:
Looks good, thanks!

PR #2782, opened 2016-11-29
fix application_helper#bulk_menu for overriding abstract_model

evanboho:
When using the helper method `bulk_menu` for a model that is not the current
`@abstract_model` the helper incorrectly gets the wording for the instance
variable instead of passing the passed variable through.


mshibuya:
Thanks!

PR #2786, opened 2016-12-03
Use ActiveSupport.on_load to correctly re-open ActiveRecord::Base.

ordinaryzelig:
Will fix https://github.com/sferik/rails_admin/issues/2785. Based on https://github.com/rails/rails/issues/23589#issuecomment-229247727.


mshibuya:
Thanks 👍

PR #2788, opened 2016-12-06
Fixed datepicker for some locales

zhengjia:
I think this is one source of the translation problem for datepicker, which manifest itself by failing to save date attribute to database.

In summary:
Double digit month should be matched first. For example, `11月` should be matched before `1月`. Otherwise a string like 2016年11月1日 would be translated to 2016年1Jan1日 because 1月 is matched first and then there is a remaining `1`.


Preen:
can this be merged?

mshibuya:
Please do not mix unrelated changes, separate them into individual PRs.

PR #2801, opened 2017-01-02
Upgrade Moment to fix DateTimePicker with Spanish

Knack:
In Spanish, month and weekday names are always written in lowercase. 

In the included version of `moment-with-locales.js` this was wrong and [names were capitalized](https://github.com/sferik/rails_admin/blob/d0e64fe0079ba60acd405c1d02220c900c1fc49e/vendor/assets/javascripts/rails_admin/moment-with-locales.js#L5126-L5140); this was causing the DateTimePicker to fail [when trying to match](https://github.com/sferik/rails_admin/blob/master/lib/rails_admin/support/datetime.rb#L43) these names with the names [from Rails I18n](https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/es.yml#L47-L58) (which are in lowercase, as expected).

This [was fixed](https://github.com/moment/moment/pull/2799) in a recent version of Moment, and now the names are in lowercase too. This Pull Request just updates the file `moment_with_locales.js` to the current version (2.17.1), and this fixes the issues with DateTimePicker and Spanish.

### DIFF
--- a/vendor/assets/javascripts/rails_admin/moment-with-locales.js
+++ b/vendor/assets/javascripts/rails_admin/moment-with-locales.js
@@ -1,2006 +1,3231 @@
-(function (global, factory) {
+;(function (global, factory) {
     typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
     typeof define === 'function' && define.amd ? define(factory) :
     global.moment = factory()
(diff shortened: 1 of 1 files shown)


sebasoga:
👍

PR #2805, opened 2017-01-08
Add config option to set locale

d-mato:
I want to use locale ':ja' but my rails app's default_locale must be ':en'. So I tried to enable to set locale for active_admin by config file.

### Usage

```ruby
# config/initializers/rails_admin.rb
RailsAdmin.config do |config|
  config.locale = :ja # I18n.locale is assigned :ja only scoped for active_admin.
 ~~~
end
```

### DIFF
--- a/spec/controllers/rails_admin/application_controller_spec.rb
+++ b/spec/controllers/rails_admin/application_controller_spec.rb
@@ -39,4 +39,12 @@ describe RailsAdmin::ApplicationController, type: :controller do
       expect(controller.send(:rails_admin_controller?)).to be true
     end
   end
+
+  describe '#_set_locale!' do
+    it 'works with config' do
+      RailsAdmin.config.locale = :fr
+      controller.send(:_set_locale!)
+      expect(I18n.locale).to eq(:fr)
+    end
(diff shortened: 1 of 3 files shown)


spec/controllers/rails_admin/application_controller_spec.rb
```diff
@@ -39,4 +39,12 @@
       expect(controller.send(:rails_admin_controller?)).to be true
     end
   end
+
+  describe '#_set_locale!' do
+    it 'works with config' do
+      RailsAdmin.config.locale = :fr
+      controller.send(:_set_locale!)
+      expect(I18n.locale).to eq(:fr)
+    end
```
mshibuya: You have to clear `RailsAdmin.config.locale`, otherwise subsequent test cases fail.

mshibuya:
This is application-specific, setting [parent_controller](https://github.com/sferik/rails_admin/blob/master/lib/rails_admin/config.rb#L63) to your custom class and changing locale there looks better.

d-mato:
Thanks for feedback! I'll try to edit parent_controller.

PR #2810, opened 2017-01-11
Loosen the dependency version of Kaminari (1.0.0)

koic:
Kaminari v1.0.0 is released.

- https://rubygems.org/gems/kaminari/versions/1.0.0
- https://github.com/kaminari/kaminari/blob/master/CHANGELOG.md#100

It seems no problem at my environment. (Ruby 2.3.3, Rails 5.0.0.1)

### DIFF
--- a/gemfiles/rails_4.0.gemfile
+++ b/gemfiles/rails_4.0.gemfile
@@ -4,6 +4,7 @@ source "https://rubygems.org"
 
 gem "appraisal", ">= 2.0"
 gem "devise", ">= 3.2"
+gem "kaminari", "~> 0.14"
 gem "rails", "~> 4.0.0"
 gem "sass-rails", "~> 4.0.3"
 gem "test-unit"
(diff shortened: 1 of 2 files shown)


koic:
The test with Rails 4.0 has failed.

https://travis-ci.org/sferik/rails_admin/builds/190896753

I propose to dropped support for Rails 4.0.

These are the reasons below.

1. [Rails 4.0 has not been maintained.](http://guides.rubyonrails.org/maintenance_policy.html)
2. [Kaminari 1.0.0 supports Rails 4.1, 4.2, 5.0, 5.1](https://github.com/kaminari/kaminari#supported-versions). In other words 4.0 is not supported.

How do you think about this?

koic:
There are other approaches. For Rails 4.0, use version of Kaminari < 1.0.0.
This approach has a small impact on existing users.

This PR has passed Travis CI with this approach.  https://github.com/sferik/rails_admin/pull/2810/commits/72f15a32eb047ef2c8ac0f3d4fa1882c33e99112

Thanks.

mshibuya:
I prefer that approach, thanks!

koic:
I get it. Thank you for the reply and merging!

PR #2817, opened 2017-01-21
Fix: filter search box duplication issue.

vrushalip:
Added condition while appending filter box.


ether1984:
it won't work for select tag in filter search box

ether1984:
var $content = $('<p id="' + field_name + '">')
                .addClass('filter form-search')
                .append('<span class="label label-info form-label"><a href="#delete" class="delete"><i class="fa fa-trash-o fa-fw icon-white"></i>' + field_label + '</a></span>')
                .append('&nbsp;' + control + '&nbsp;' + (additional_control || ''));
if ($('p#' + field_name).length == 0) {
    $('#filters_box').append($content);
}

evanboho:
I did not see this PR and fixed the same bug in https://github.com/sferik/rails_admin/pull/2998

mshibuya:
Fixed by #2998.

PR #2824, opened 2017-01-28
Handle datetime after switching locales properly

roberthead:
When switching locales, our datetime fields were not displaying in the proper format. Removing the memo-ization of @parser in the Datetime field type fixes the problem.

Would be happy to pair with a maintainer or contributor who knows the codebase and project spec style better.


mshibuya:
Thanks!

PR #2832, opened 2017-02-09
Add ability to show papertrail models

petrovmp89:
Users are can't show paper_trail models in UI with default names, because they were added in excluded models.


mshibuya:
Causing SyntaxError.

PR #2839, opened 2017-02-24
Use jQuery2 

MatthiasRMS:
As I said in this [issue](https://github.com/sferik/rails_admin/issues/2838), jQuery 3 is causing the gem to break.

I ran the tests and it seems like reverting to jQuery2 does not break anything.

### DIFF
--- a/app/assets/javascripts/rails_admin/rails_admin.js
+++ b/app/assets/javascripts/rails_admin/rails_admin.js
@@ -1,4 +1,4 @@
-//=  require 'jquery'
+//=  require 'jquery2'
 //=  require 'jquery_ujs'
 //=  require 'jquery.remotipart'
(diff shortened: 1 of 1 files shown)


mshibuya:
Can you make Rails 4.0/4.1 builds pass?

MatthiasRMS:
I ran` bundle exec rake default`, does it run only for Rails 5 ? 
How can I specify to run it for Rails 4.0 ? 

When running `bundle exec rake default`, only error came from `PaperTrail` but this error is still present when sticking with jQuery 3 so I don't think it comes from this PR.

mshibuya:
Closing because of this: https://github.com/sferik/rails_admin/issues/2838#issuecomment-401527294

Issue #2773, opened 2016-11-18
Root actions on the sidebar. Not on dashboard and top nav 


Kalagan:
Is there a way to show custom pages on the sidebar on the left?
Without having a specific Model for this page/action

mshibuya:
Closed by #2844.

Issue #2345, opened 2015-06-24
New navigation DSL


mshibuya:
Current navigation configuration(`navigation_static_links` and `navigation_static_label`) is ad-hoc and provides little flexibility. I'd like to replace them with `register_instance_option`-style DSL.

PR #2844, opened 2017-03-06
customizable menu

zambot:
Gives custom actions the ability not to show up in the topnavigation, but below custom labels in the sidenavigation.
This way the navigation is more intuitive, if you're using custom actions excessively.

solves #2773 
extends #2345


mshibuya:
Thanks, I'll write the specs.

jacobjlevine:
From the description, this seems like it could have multiple uses, but I'm having a bit of trouble envisioning exactly what this entails. Has this functionality been documented in the wiki?

PR #2846, opened 2017-03-06
Find the last created record by `created_at` field, not primary key

chumakoff:
The `.last` method by default finds the last record ordered by primary key.
The `created_at` column and a primary key column don't always come in the same order.
So, in order to find the last created record we should order explicitly by `created_at` column.

Since we only need to get the `created_at` value, there is no need to load and instantiate the whole record.
We can fetch just `created at` value using `.pluck` method.


mshibuya:
We can't make this default, because users don't always have `created_at` column indexed.

PR #2857, opened 2017-03-23
Set global associated_collection_cache_all to false

shreyas123:
I have a large tables and having a global scope to return false for associated_collection_cache_all instead of checking the count and returning false is a large time consuming for me. 

Add a config called 'associated_collection_cache' where it is default true and when provided false will not query the count in 'associated_collection_cache_all'.


mshibuya:
Just doing `associated_collection_cache_all { false }` in field config is enough.

PR #2860, opened 2017-03-25
Add instance option to append a CSS class for rows

MrOutis:
Follow up of PR #2606 , with tests
-------------------------------------------

As Rails Admin uses Bootstrap, I wanted to use the table's contextual classes in my lists.
This is an approach to implement that using a method declared in RailsAdmin::Config,
it has access to the object being rendered into the row.
### Usage

``` ruby
RailsAdmin.config do |config|
config.model Event do
  list do
    row_css_class do
      bindings[:object].mode == 'panic' ? 'danger' : 'success'
    end
end
```

### DIFF
--- a/app/views/rails_admin/main/index.html.haml
+++ b/app/views/rails_admin/main/index.html.haml
@@ -90,7 +90,7 @@
           %th.last.shrink
       %tbody
         - @objects.each do |object|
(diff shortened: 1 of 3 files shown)


mshibuya:
Thanks 👍

PR #2862, opened 2017-03-29
Use phantomjs2

waghanza:
`Trusty` environment on `travis` has **phantomjs** `2.0.0` instead of `1.9.3` in `precise`

@see https://github.com/travis-ci/travis-ci/issues/3225#issuecomment-151367517

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,8 @@
-dist: trusty
-
 language: ruby
 
 before_install: gem install bundler
 
 before_script:
-  - phantomjs --version
   - cd spec/dummy_app
   - bundle exec rake rails_admin:prepare_ci_env db:create db:migrate
   - cd ../../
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -29,6 +29,8 @@ gemfile:
 
 services:
   - mongodb
+  - mysql
+  - postgresql
```
mshibuya: Looks unrelated to PhantomJS, is this correct?

waghanza: nop, in fact on travis  `precise` **env** (default) those services are loaded by default, but with `trusty` env (to have phantom2) we must list all service that have to be started

.travis.yml
```diff
@@ -1,11 +1,8 @@
-dist: trusty
-
```
mshibuya: You want to switch to `trusty`, why is removed?

waghanza: You're write. Kind a mess

waghanza:
Closed, and re-opened #2873

PR #2865, opened 2017-04-01
PaperTrail + Inheritance

lazebny:
This PR resolves issues https://github.com/sferik/rails_admin/issues/1524, https://github.com/sferik/rails_admin/issues/2332, https://github.com/sferik/rails_admin/issues/2691
It is based on https://github.com/sferik/rails_admin/pull/1567/files.

PaperTrail stores a base_class for  each subclass, so we can't fetch versions with searching by subclass. STI model allows us to fetch records with query condition which includes STI type, so `model.model.all` should return a valid AR relation for current STI type and we can fetch a versions by this relation. 

When we have a deal with object versions think we can't do better than to use native PaperTrail method versions, so `object.versions` should work well.  

These changes were checked on our production db.

Here is illustration how PaperTrail works:

> class PaperTrailTest; end
class PaperTrailTestSubclass < ::PaperTrailTest; end
class PaperTrailTest::SubclassInNamespace < PaperTrailTest; end
versions = PaperTrail::Version.where(item_type: 'PaperTrailTest')
PaperTrailTest.create 
versions.count # 1
PaperTrailTestSubclass.create 
versions.count # 2
PaperTrailTest::SubclassInNamespace.create
versions.count # 3

### DIFF
--- a/lib/rails_admin/extensions/paper_trail/auditing_adapter.rb
+++ b/lib/rails_admin/extensions/paper_trail/auditing_adapter.rb
@@ -97,12 +97,9 @@ module RailsAdmin
             sort_reverse = 'true'
           end
 
-          model_name = model.model.name
-
           current_page = page.presence || '1'
 
-          versions = version_class_for(model_name).where item_type: model_name
-          versions = versions.where item_id: object.id if object
+          versions = object.nil? ? versions_for_model(model) : object.versions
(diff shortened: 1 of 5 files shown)


v-shmyhlo:
:+1, i have same issue, would be nice to have these changes in master

mshibuya:
Excellent, thanks!

PR #2868, opened 2017-04-16
show warning if field has 'dependent: :destroy' option

marat-chardymov:
When `dependent: :destroy` option is set, we show two windows with multiselect, but it's not obvious that orphan records might be deleted(when admin detach records from parent with _ra-multiselect-item-remove_). I've added warning for this case 
https://github.com/sferik/rails_admin/issues/2762 

P.S. Sorry for creating PR without tests: I haven't managed to run them locally because of 'Please set up Papertrail's version model explicitly. ' error, so I've created PR to run tests on CI first


mshibuya:
Help option can be used if you want to specifically warn users.
https://github.com/sferik/rails_admin/wiki/Fields#help

PR #2879, opened 2017-05-18
IrreversibleOrderError patch

kevTheDev:
I had an issue where one of my models was using a default sort scope with NULLS LAST

Since upgrading to Rails 5.1.1 I got IrreversibleOrderError raised on loading rails_admin - by loading I mean visiting the /admin route.

Not sure of any other side effects from this but for me at least it means rails admin no longer just falls over with a 500 - and if this doesn't get merged it can at least serve as a heads up

Thanks

Kev


mshibuya:
Turning dashboard statistics off will work for your case, If you don't need them.
https://github.com/sferik/rails_admin/wiki/Dashboard-action#disabling-record-count-bars

kevTheDev:
Thanks for your help - that does work

But this is not really a fix and therefore still what I would consider to be an open issue

PR #2887, opened 2017-05-30
version bump

cpalafox:
There are some changes made, mainly the removal of `alias_chain_method` that throw some deprecation warnings in Rails 5, that are not being applied due to the lack of a version bump.


mshibuya:
Just shipped 1.2.0, thanks.

PR #2890, opened 2017-06-03
fix HTML of filter refresh on history page

sethjeffery:
Provides a fix for https://github.com/sferik/rails_admin/issues/2889 so that the filter button looks correct again. Same approach as https://github.com/sferik/rails_admin/pull/2706

image


mshibuya:
Thanks!

jschwindt:
I have this error again in version 2.0 with Rails 6.0
Screen Shot 2019-09-13 at 16 34 30
and the solution was to exchange single with double quotes in data-disable-with that contains html tags as shown in my PR.

PR #2898, opened 2017-06-28
fix: save_and_edit should authorize the current object

bf4:
was authorizing the class. 

Not sure why I'm the first person to see this. Looks old I found this https://github.com/sferik/rails_admin/commit/4ec9719920fb23019ce65b417ff5c4463a9b5e3f#commitcomment-22816721

I was kind of surprised my `edit?` policy (using Pundit) was called twice.  The first time it was authorizing the instance and the second the class.  It might be better to pass in a local to the partial since whether or not we can edit has already been determined

### DIFF
--- a/app/views/rails_admin/main/_submit_buttons.html.haml
+++ b/app/views/rails_admin/main/_submit_buttons.html.haml
@@ -8,7 +8,7 @@
       - if authorized? :new, @abstract_model
         %button.btn.btn-info{type: "submit", name: "_add_another", :'data-disable-with' => t("admin.form.save_and_add_another")}
           = t("admin.form.save_and_add_another")
-      - if authorized? :edit, @abstract_model
+      - if authorized? :edit, @abstract_model, @object
         %button.btn.btn-info{type: "submit", name: "_add_edit", :'data-disable-with' => t("admin.form.save_and_edit")}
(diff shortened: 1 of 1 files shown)


bf4:
failures are in jruby, unrelated 

```
 bundle exec rake rails_admin:prepare_ci_env db:create db:migrate

jruby: warning: unknown property jruby.cext.enabled


rake aborted!
Bundler::GemRequireError: There was an error while trying to load the gem 'capybara'.
Gem Load Error is: undefined method `call' for nil:NilClass
Did you mean?  caller
Backtrace for gem load error is:
org/jruby/RubyMarshal.java:145:in `load'
```

I can submit a fix to the JRUBY_OPTS

mshibuya:
Thanks for the fix, but it would be great if you could add a test.

bf4:
Sure, will do! Thanks for getting back

mshibuya:
After thinking carefully I decided not to merge this, because it doesn't make sense to decide if a object is editable or not based on its attributes **before** submitting form(the attributes may change during the edit!). At least, this is what the change in https://github.com/sferik/rails_admin/commit/4616200f79ce587fd144257b663f0cda6307c31e intended.

Thanks for contribution anyway.

bf4:
ok :)

PR #2905, opened 2017-07-21
Fix #2398: Broken Client Side Validation for filtering select

### LINKED ISSUE #2398: Broken Client Side Validation for filtering select
Hi,

when a required attribute is left empty, upon form submission, a tool-tip is shown on the field.

When the required attribute is a relation, the client-side navigation **does not work**.

In Chrome there is no visible reaction, but in the console the following error pops up:
`An invalid form control with name='property[feature_id]' is not focusable.`

In Firefox, the tooltip shows up in a completely wrong location (see picture below).

I'm pretty sure the the cause is that for filtering-select, the **invisible select tag** has the attribute
`required="required"`
and this freaks out Chrome and breaks the positioning of the tooltip on Firefox.

I believe that for associations, the required attribute needs to be moved to the visible controls.

screen_shot_2015-09-10_at_15_28_02

I'd gladly contribute a pull-requrest myself, but I don't understand very much the form-rendering logic in rails-admin.

williamherry:


bpicolo:
@mshibuya  this would be a very useful fix to have in. Validations are just totally busted for these autocomplete dropdowns right now

mshibuya:
Sorry for delay, thanks!

PR #2908, opened 2017-07-28
Fix for translation filters

nicocedron:
The JSON of JS translations was passed as text, so the translation was never found.
The solution was to add parse to the text to convert it to JSON.


mshibuya:
Merged #2925.

PR #2909, opened 2017-07-31
Added few and zero, many

Albina-art:
Adding declines.


mshibuya:
This looks too specific to your usecase, please find a solution for pluralization that works globally, including your locale.

PR #2910, opened 2017-08-02
Beautify JSON datatype

dalpo:
Beautify JSON values into the show page.

Before:
screenshot from 2017-08-02 10-08-50

After:
screenshot from 2017-08-02 10-08-20


mshibuya:
Overlaps with #2937, please update if further improvement exists.

dalpo:
Ok, thanks @mshibuya!

Issue #2914, opened 2017-08-08
Invalid SQL syntax generated for M2M associations tables


amarchenkotruecar:
There're 3 tables: users(id, ...), user_groups (user_id, group_id), groups(id, name, ...). 

UserGroup class has:
```
 belongs_to :user
 belongs_to :group
```

User class has: 
```
has_many :user_groups
has_many :groups, through: :user_groups
```

Group class has:
```
has_many :user_groups
has_many :users, through: :user_groups
```

In other words, user_groups is a typically setup M2M association table represented in rails via has_many :through on both sides of the association. This works flawlessly with all our code base but starting recently it started breaking rails_admin while opening a User record or navigating to a list of UserGroups with this with error: `PG::UndefinedColumn: ERROR: column user_groups.desc does not exist`. We have also seen the same error with other association tables that are set up in exact the same way. 

Just in case that matters, the tables are namespaced (in the same schema but not in default one). We use papertrail with association tracking (but not for the tables mentioned above so it shouldn't interfere). 

```

File "/usr/local/bundle/gems/activerecord-5.0.4/lib/active_record/connection_adapters/postgresql_adapter.rb" line 598 in async_exec
2
File "/usr/local/bundle/gems/activerecord-5.0.4/lib/active_record/connection_adapters/postgresql_adapter.rb" line 598 in block in exec_no_cache
3
File "/usr/local/bundle/gems/activerecord-5.0.4/lib/active_record/connection_adapters/abstract_adapter.rb" line 590 in block in log
4
File "/usr/local/bundle/gems/activesupport-5.0.4/lib/active_support/notifications/instrumenter.rb" line 21 in instrument
5
File "/usr/local/bundle/gems/activerecord-5.0.4/lib/active_record/connection_adapters/abstract_adapter.rb" line 583 in log
6
File "/usr/local/bundle/gems/activerecord-5.0.4/lib/active_record/connection_adapters/postgresql_adapter.rb" line 598 in exec_no_cache
7
File "/usr/local/bundle/gems/activerecord-5.0.4/lib/active_record/connection_adapters/postgresql_adapter.rb" line 587 in execute_and_clear
8
File "/usr/local/bundle/gems/activerecord-5.0.4/lib/active_record/connection_adapters/postgresql/database_statements.rb" line 103 in exec_query
9
File "/usr/local/bundle/gems/activerecord-5.0.4/lib/active_record/connection_adapters/abstract/database_statements.rb" line 373 in select
10
[...]

PR #2915, opened 2017-08-15
raise MissingPrimaryKey

chopraanmol1:
Fix for #2914 

Model without a primary key will flash message


mshibuya:
Showing a message in runtime seems inadequate, how about raising the exception on initialization?

chopraanmol1:
@mshibuya 
Rails lazily loads model's schema detail from database. Raising an exception during initialization will slow down rails server boot.

mshibuya:
Telling end-users that they have missing primary key is useless, because the issue must be resolved by devs anyway.
Just exclude the model if don't want it to show up and cause SQL error.

PR #2917, opened 2017-08-22
Toggle checkboxes

clurdish:
This adds the ability to toggle the checkboxes on the list section. The bulk menu is also disabled whenever the checkboxes instance variable is set to false on the list section. 

The reason for this feature is to remove some of the clutter on the list page if no bulk actions are being used. Some of our clients have said it's distracting without bulk actions being available. 

All tests are passing on CentOS 7 with Ruby 2.3.3. Code coverage has increased.


mshibuya:
Looks good, thanks!

PR #2925, opened 2017-09-11
Fix JS translations not be object but string

zakihaya:
I want to use I18n locale setting but yaml file didn't work.
Because `translations` value become string.

So fix to become `translations` hash.


mshibuya:
Thanks.

PR #2937, opened 2017-09-29
Wrap json field in pre block

lazebny:
Resolves issue with json formatting in view https://github.com/sferik/rails_admin/issues/2872.

Rails Admin json api works correctly with this addition and returns plain json.


mshibuya:
Thanks!

PR #2943, opened 2017-10-17
Fix Fields::Types::Time timezone issue

dmedvinsky:
When Field::Time parses input, it converts the Time to current timezone,
but then throws away the zone info when converting the Time to string
via strftime, so the created DateTime is parsed without the zone
information, thus offsetting the input value to the zone offset hours.


mshibuya:
Please add some specs to reproduce the issue.

mshibuya:
Please reopen on update.

PR #2948, opened 2017-10-22
Create rails_admin.pt-BR.yml

eliasnogueira:
Add a pt-BR locale


mshibuya:
Thanks for the PR, but we maintain all locales in Wiki.

Issue #1235, opened 2012-07-03
Other custom text editors? (markdown via MarkItUp)


Tuckie:
After seeing the codemirror pull: https://github.com/sferik/rails_admin/pull/1164 I was wondering how difficult it would be to add other client side editors into rails admin? 

I'm looking at incorporating markdown support into my application, but a little extra assistance via MarkItUp is probably going to  be needed.

https://www.ruby-toolbox.com/gems/markitup_rails3
http://markitup.jaysalvat.com/home/

Looking at code changes that was required, it doesn't appear to be very modular.  Is there a better way to cleanly integrate different editors (maybe just as a markdown field type)?

Ryo10Leo:
hhhh****

jjj__

PR #2949, opened 2017-10-24
Add markdown field support through SimpleMDE

d9su:
I was looking for markdown editing support in `rails_admin` and came across this stale issue #1235. After some digging I couldn't find a satisfactory solution in built-in support or plugins, so thought I'd take a stab at it myself.

I decided to go with [SimpleMDE](https://github.com/sparksuite/simplemde-markdown-editor) after some research since it looks neat and very simple to use. 

With this feature, you can do `field :text_field, :simple_mde` and have a SimpleMDE markdown editor. The easiest way to enable it is [`simplemde-rails`](https://github.com/KHN190/simplemde-rails) which unfortunately uses a somewhat outdated SimpleMDE version, ~~but I've opened an issue there to hopefully update it~~ fixed in https://github.com/KHN190/simplemde-rails/issues/1. Users can also just load SimpleMDE themselves.

**This is how it looks like:**

screen shot 2017-10-24 at 3 24 17 pm

**Test coverage:**

-users-rexsu-src-github com-rails_admin-coverage-index html

Linter is complaining that 2 lines in the coffeescripts are too long... but there are longer lines existing in the same file, not sure if I should bother with it.

cc @mshibuya


d9su:
Hey hey, may I get some thoughts on this? It'd be great if we can get it in the gem, so let me know what I can do to achieve that 😃

coorasse:
Would be really glad to see this feature 👍

philschoefer:
This would be a great addition. Please merge!

mshibuya:
Nice work, thanks!

PR #2952, opened 2017-10-31
Fix filtering list by enum integer value

smathieu:
This correctly parses and decode the enum values for filtering in the
list view.

Note: I also had to comment out a test that meses up the internal state
of ActiveRecord. I looked at the AR source and I don't think there's a
way of preventing it. Maybe this test should be rewritten.


smathieu:
@mshibuya any chance we can merge this? We're running this in our production environment without issue.

mshibuya:
Closed in favor of #2971.

PR #2968, opened 2017-12-10
Add big table limited pagination

lazebny:
### Overview

Added limited pagination 'prev', 'next' buttons only. It allows not to do `count(*)` which is slowly on big tables.

P.S. If this approach is ok we can reuse it in  https://github.com/sferik/rails_admin/issues/2808

### DIFF
--- a/app/views/kaminari/ra-twitter-bootstrap/without_count/_paginator.html.haml
+++ b/app/views/kaminari/ra-twitter-bootstrap/without_count/_paginator.html.haml
@@ -1,0 +1,4 @@
+= paginator.render do
+  %ul.pagination
+    = prev_page_tag if !current_page.first?
+    = next_page_tag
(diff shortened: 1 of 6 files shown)


lazebny:
@mshibuya could you check it?

mshibuya:
Thanks for the PR, I like this feature.

app/views/kaminari/ra-twitter-bootstrap/without_count/_paginator.html.haml
```diff
@@ -0,0 +1,4 @@
+= paginator.render do
+  %ul.pagination
+    = prev_page_tag if !current_page.first?
+    = next_page_tag # if !current_page.last? && !current_page.out_of_range?
```
mshibuya: This commented-out code should be removed.

lazebny: Thanks, will do.

lazebny:
@mshibuya comments are removed, please let me know if need to fix something else.

mshibuya:
Merged in, thanks!

PR #2971, opened 2017-12-24
Fixed filter by enum filter when label <> value

mshytikov:
Refactored code to remove unnecessary calls to `parse_value`.
Which were affecting filter by enum (rails >= 5.0):
If enum defined like
```
enum size: { L: 'l', S: 's' }
```
Due to multiple calls, the chain when filtering by size is following:
```
parse_value("l") => "L"
parse_value("L") => nil
```
Which leads to non working filtering.
NOTE: `parse_value` returns correct result in both cases above

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gem 'devise'
 group :active_record do
   platforms :ruby, :mswin, :mingw do
     gem 'mysql2', '~> 0.3.14'
-    gem 'pg', '>= 0.14'
+    gem 'pg', "~> 0.21"
     gem 'sqlite3', '>= 1.3'
   end
 end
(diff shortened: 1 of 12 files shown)


mshibuya:
Could you add a test case for that issue?

mshytikov:
Sure, I will add tests.
(need to find some time, should be ready in few days. )

mshytikov:
work in progress.

mshytikov:
Turns out that fix provided in #2952 for `active_record_enum.rb` is also required to pass the build for some older rails versions

mshytikov:
@mshibuya I have added tests. And have rebased them, so it is possible to see in travis logs that without fix they are failing. 
I was quite unlucky with this PR that I also faced `pg` update see the last commit message. 
Could you please review PR ?

mshibuya:
Merged in 857696119c70e6603fa7de0dca8b983ba01c5195, thanks so much!

mshytikov:
👍  thank you.

PR #2973, opened 2017-12-30
Fix json field tag <pre> in edit form

lazebny:
### Overview

Fixed redundant tag `pre` in edit form.


arnvald:
@lazebny thank you for such a quick response and fix! I've just checked this patch in my application and it works fine - the `show` action displays json wrapped in `<pre>` tag, but the `edit` action doesn't add it anymore 👍

mshibuya:
Thanks for the fix.

PR #2974, opened 2017-12-30
Add association list action

lazebny:
### Overview

Model associations can be viewed inside a model layout with the same features as on index page. 

@mshibuya want to hear your feedback about this feature, if it is positive i'll complete PR with more tests and resolve mongoid issue.


mshibuya:
Sorry for long unresponsiveness, but this feels heavy.
I think this can be more integrated with index action, like implementing special association-scoped mode in index.

PR #2980, opened 2018-01-15
Fixed displaying empty json in edit form

arnvald:
## The problem

I have empty json saved in database (not `null`, just `{}`)
I want to edit my record in RailsAdmin

**What I see**: the json field is empty
**What I should see**: field should display `{}` instead

## The solution

Currently we check whether to display json based on `value.present?`, which returns `false` for empty json. Instead we should just check `value`


mshibuya:
👍

PR #2983, opened 2018-01-18
use value instead url in file's link_to body 

buncismamen:
use filename/value instead full url in file's link_to body


mshibuya:
Could you write a spec for this?

buncismamen:
@mshibuya I want to write the spec, but I don't know how since it have no spec before, can you guide me how to write the spec? thank you

mshibuya:
Cherry-picked as e3770baf485b6d3dc56a33b35973339785fcb374.

Issue #1819, opened 2013-10-29
"Is Exactly" filter still uses LIKE


ehoch:
Why does is the "Is Exactly" filter using LIKE? This removes any ability to rely on indexes in the list view. Some of my tables have millions of rows making a like of that nature impossible...

kevin12gigs:
This is still an issue with the latest version 0.8.1. As @ehoch mentioned this is **horribly** inefficient, especially for indexed columns. I have put a small patch in that makes it use the '=' operator, and we saw a huge improvement in performance for "Is Exactly" Filters. I simply added this to build_statement_for_string_or_text:

rihardssceredins:
I actually see that when using "is exactly" filter the query built with ILIKE.

rihardssceredins:
When could this be fixed?

mshibuya:
Fixed by #3000.

PR #2988, opened 2018-01-25
don't use like for 'is_exactly' queries

sliminas:
This PR solves issue #1819 so that when filtering for string fields the LIKE operator is not used anymore if the filter is 'is_exactly'.


evanboho:
Yes! But your specs are failing. I made a similar PR with passing specs:
https://github.com/sferik/rails_admin/pull/3000

PR #2989, opened 2018-02-02
Allow to customize forgery protection settings

arnvald:
Currently `protect_from_forgery` is always called with `with: :exception`,
which might becausing problems in certain applications.

This commit enables configuration for `protect_from_forgery` method,
and keeps `with: :exception` as default settings


mshibuya:
Nice, thanks!

sarink:
Is there anyway to disable it completely? How would one use `skip_forgery_protection`? https://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection/ClassMethods.html

PR #2998, opened 2018-03-07
Avoid duplication of filters on back

evanboho:
Bug: filters are duplicated when using back button
Repro steps:
Using Chrome, visit a list page, apply a filter, click into an element, click back button
Using these steps, I see a duplicated filter. RailsAdmin is appending the filters regardless
of whether or not they are there.

This fixes that by adding an id to each filter based on the name and index.
Before appending, it removes any filters with that id.

I haven't been able to figure out how to write specs for this. Any assistance would be appreciated.


mshibuya:
Great, thanks!

PR #2999, opened 2018-03-07
Fix "can't modify frozen array" exception under certain conditions

glebtv:
This fixes "can't modify frozen array" exception which occurs for me in development in many real-world apps.

It happens when @_fields is returned from HasFields#fields and then it is being frozen when assigning it to @_ro_fields.

The spec included is as close as I could get it to my use-case (real app does not need to add a new field type, any change to code that causes my app to reload triggers this exception)


fmh:
👍

mshibuya:
Looks good, thanks!

PR #3000, opened 2018-03-07
Use exact match for 'is' and '='

evanboho:
This is in response to #1819 and is inspired by #2988
This changes less code than #2988, which is currently not
passing CI, and preserves postgres checks.


sliminas:
Great job :+1:

nogates:
I was going to suggest doing something like this. Using `ILIKE` is problematic with large data sets, and makes the `queryable` field unusable.
:+1: to get this merged

rihardssceredins:
When could this be merged? The list view is completly usuable for filtering string fields..

linh-yesware:
@mshibuya May we ask what's blocking this pull from being merged? How can we help get it merged? Thanks!

vincentwoo:
Could we have a gem cut with this change?

rihardssceredins:
@mshibuya is a release planned anytime soon? This dramatically affects usage for large tables.

vincentwoo:
@sferik @mshibuya it would really, really, really, really save our asses if we could get a gem cut with this commit in it. Right now we basically cant use rails admin for any large table.

rihardssceredins:
@mshibuya when a next release is planned? Why the delay? This is a serious issue for larger apps. :)

Issue #2919, opened 2017-08-28
rails_admin throw ActionController::InvalidAuthenticityToken in rails-api 


buncismamen:
rails_admin throw ActionController::InvalidAuthenticityToken when I want to create model

![image](https://user-images.githubusercontent.com/14258779/29773791-283d1a8c-8c29-11e7-8cde-b9eea88c9b28.png)

rails version 5.1.2 with --api mode
rails_admin version 1.2.0

change `config.api_only` from `true` to `false` in `application.rb` fix the problem idk why, anyone care to explain?

buncismamen:
I think its related with this https://github.com/sferik/rails_admin/issues/2830
adding these stack into rails middleware also fix this 
```
# application.rb
config.session_store :cookie_store
config.middleware.use ActionDispatch::Cookies
config.middleware.use ActionDispatch::Session::CookieStore, config.session_options
```
https://github.com/sferik/rails_admin/issues/2830#issuecomment-323622174

efrapp:
Thanks @buncismamen it worked for me!!

visheshruparelia:
Thank you, this fixed it. But I don't get why this worked, can you explain? @buncismamen

buncismamen:
@visheshruparelia because rails-api remove some middleware stacks that rails admin needs.
so you just need to re-added it.

mshibuya:
Fixed by #3006.

1054949000:
<img width="1740" alt="截屏2022-04-08 下午6 42 28" src="https://user-images.githubusercontent.com/6510093/162420438-90040ba7-4f5c-44d2-9294-cf49929b61b1.png">
<img width="697" alt="截屏2022-04-08 下午6 42 49" src="https://user-images.githubusercontent.com/6510093/162420453-9d70fbdd-4ad7-465f-86ad-56bebc98ee4a.png">


i have the problem: back 422
ActionController::InvalidAuthenticityToken

i have add the code in application.rb

config.session_store :cookie_store
    config.middleware.use ActionDispatch::Cookies
    config.middleware.use ActionDispatch::Session::CookieStore, config.session_options
    config.middleware.use Rack::MethodOverride
    config.api_only = false

PR #3006, opened 2018-03-25
fixes issue #2919

robbywashere:
Add missing cookie middleware when rails is used in api mode

I am not entirely sure if this is the correct way to go about this. My rails engine knowledge is limited. Will this also add these middlewares to the main application or _just_ this engine (rails_admin) ?


mshibuya:
Merged in, thanks!

ritesh2741:
@mshibuya Adding the cookie middleware overwrote the code that I had in config/initializers/session_store.rb. I had my own cookie name but it defaulted back to '_session_id' . I am using Rails 4.2.10. This is a fix for Rails 5 but i believe it caused problem for me in Rails 4 .  Quick solution was adding the 'middleware config' to application.rb, but the _session_id was persistent along with my custom cookie.

PR #3007, opened 2018-03-28
Fix incorrectly comparing numeric columns with empty string when handling blank operator

### LINKED ISSUE #2995: Filtering records with association is blank generates invalid SQL
## Env
`rails_admin 1.3.0`

## Filters selected
screen shot 2018-03-05 at 1 53 44 pm

## Query params
screen shot 2018-03-05 at 1 54 04 pm

## SQL
```sql
SELECT  "buildings".* FROM "buildings" WHERE ((buildings.name_en ILIKE 'Name')) AND ((buildings.original_building_id IS NULL OR buildings.original_building_id = '')) ORDER BY buildings.id desc LIMIT $1 OFFSET $2
```

PikachuEXE:
Fix #2995

### DIFF
--- a/lib/rails_admin/adapters/active_record.rb
+++ b/lib/rails_admin/adapters/active_record.rb
@@ -155,6 +155,8 @@ module RailsAdmin
           case @type
           when :boolean
             boolean_unary_operators
+          when :integer, :decimal, :float
+            numeric_unary_operators
           else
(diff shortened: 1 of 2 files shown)


PikachuEXE:
Hello~?

mshibuya:
Looks nice, thanks!

Issue #3014, opened 2018-04-19
Rails Admin would not run with Rails 5.2 (Could not find table 'active_storage_attachments')


rajahafify:
Rails Admin would not run with Rails 5.2 without running `bin/rails active_storage:install` and `bin/rails db:migrate` first. This is because of the introduction of active_storage which in my limited understanding creates a connection to the DB which is used by RailsAdmin to bootstrap the Dashboard.

Rails Core Team thinks that this is OK [https://github.com/rails/rails/pull/31315](https://github.com/rails/rails/pull/31315) and their solutions is basically, to show an appropriate error message and the appropriate workaround, when a user face this issue, similar to how Rails currently handled a pending migration.

On one had, I think this is problematic, since its different from what it used to be. I have to read ActiveStorage documentation to find out how to fix the issue. Maybe having this mentioned in Readme would have been useful.

On the other hand, this problem is exacerbated by RailsAdmin not showing the normal error message that is usually provided by Rails raised in a different issue [https://github.com/sferik/rails_admin/issues/3012](https://github.com/sferik/rails_admin/issues/3012)

This actually amplifies the difficulty of debugging the issue since we couldn't have seen the workaround provided by the Rails Core Team.

I've included a sample project that has the problem you guys can test it out your self.
[https://github.com/rajahafify/rails_admin_active_storage_issue](https://github.com/rajahafify/rails_admin_active_storage_issue)

Detail
[Error Logs](https://gist.github.com/rajahafify/19595f5d7b21bddfdce0d9e64da8dbd7)

Ruby Version
2.3.4
Also tried with 2.5.0

RailsAdmin Version
1.3

Tested on my Mac and a Ubuntu VM

abelmartin:
Exactly what I needed.  Thanks!

ziaulrehman40:
@dmilisic thanks, its strange though, i was using following syntax:
```
  config.excluded_models << ActiveStorage::Blob
  config.excluded_models << ActiveStorage::Attachment
```

And my app won't load and complain `has_one_attached` not defined.
Doing it your way fixed the issue.

PR #3015, opened 2018-04-19
Possible fix for issue #3014

rajahafify:
Updated the Readme to notify user who's using Rails 5.2 on how to generate ActiveSupport table migration file and migrating it.


bodrovis:
It should be `rails active_storage:install`

rajahafify:
Yes, i've updated the commit yesterday to fix the typo

Issue #2272, opened 2015-04-27
Horizontally scrolling tables in list view - pull request potential?


sbull:
I recently worked out a way to make the "list" view tables more useful to me - showing more/all columns with horizontal scrolling and frozen header columns, as I described here: https://github.com/sferik/rails_admin/wiki/Horizontally-scrolling-table-with-frozen-columns-in-list-view
It's mostly done in javascript, but could be better suited to be integrated in the gem instead.

I'm wondering if there would be support for making this a configuration option. If the feature would likely be accepted, I'll work on a pull request for it. I'm sure I'll need some guidance along the way - such as naming for the config option, how to build a test case, etc. Should I take a crack at this, or do you think it's better to live in userland customization instead of in the gem?

sbull:
@mshibuya do you have thoughts on this? Or should I discuss it with someone else? Thanks!

iMacTia:
It should definitely be included as an option for the gem, would make a lot of sense!

SuperMasterBlasterLaser:
@sbull I have used your javascript code, and that kind of result I have got:

![1](https://cloud.githubusercontent.com/assets/8684688/11365296/9b9f3492-92cf-11e5-8aed-4662a8e87fff.PNG)

It "destroys" table.

sbull:
I've slightly updated the JS with rails_admin 0.6.8. Here's what I get (rails_admin 0.6.8, Chrome 46.0).

![railsadminhorizscroll](https://cloud.githubusercontent.com/assets/1115369/11372389/cf7cd36a-9292-11e5-9e00-ea7e3374c090.jpg)

martnst:
+1 on including it with rails_admin per default

opensourceame:
+1 for including it with Rails but I think it should be optional in the config. The JS should also add some classes to the wrapper elements so it's easily to customise the look.

twikah:
+1 on including it with rails_admin as optional in the config!

mshibuya:
#3017 was merged in.

PR #3017, opened 2018-04-29
Add config.horizontal_scroll_list to enable horizontal scrolling colu…

sbull:
…mns and frozen row headers instead of paginated sets

Addresses Issue #2272.

### DIFF
--- a/app/views/rails_admin/main/index.html.haml
+++ b/app/views/rails_admin/main/index.html.haml
@@ -51,43 +51,11 @@
   jQuery(function($) {
   #{ordered_filter_string}
   });
-- if horiz_scroll_enabled && horiz_scroll_frozen_cols > 0
-  :javascript
-    (function(){
-      if (window.rails_admin_horizontal_scroll_list) { return; } // Don't add this handler multiple times.
-      window.rails_admin_horizontal_scroll_list = true;
-      var $ = jQuery;
-      var setFrozenColPositions = function(){
-        var firstPosition, $td;
-        $('#bulk_form').find('table tr').each(function(index, tr){
-          $(tr).find('.scroll-frozen').each(function(idx, td){
-            $td = $(td);
-            if (idx === 0) {
-              firstPosition = $td.position().left;
-            }
-            td.style.left = ($td.position().left-firstPosition)+'px';
-          });
-        });
-      };
-      $(window).on('load', setFrozenColPositions); // Update after link icons load.
-      $(document).on('rails_admin.dom_ready', setFrozenColPositions);
-    }());
 
 %style
   - properties.select{ |p| p.column_width.present? }.each do |property|
     = "#list th.#{property.css_class} { width: #{property.column_width}px; min-width: #{property.column_width}px; }"
     = "#list td.#{property.css_class} { max-width: #{property.column_width}px; }"
-  - if horiz_scroll_enabled
-    \.table-wrapper { margin-bottom: 20px; overflow-x: auto; }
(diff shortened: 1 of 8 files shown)


app/views/rails_admin/main/index.html.haml
```diff
@@ -39,11 +51,43 @@
...
+            if (idx === 0) {
+              firstPosition = $td.position().left;
+            }
+            td.style.left = ($td.position().left-firstPosition)+'px';
+          });
+        });
+      };
+      $(window).on('load', setFrozenColPositions); // Update after link icons load.
+      $(document).on('rails_admin.dom_ready', setFrozenColPositions);
+    }());
```
mshibuya: Place your javascripts and stylesheets in appropriate location, not putting them into the view.

lib/rails_admin/config.rb
```diff
@@ -59,6 +59,9 @@ class << self
       # Set the max width of columns in list view before a new set is created
       attr_accessor :total_columns_width
 
+      # Enable horizontal-scroll table in list view, ignore total_columns_width
+      attr_accessor :horizontal_scroll_list
```
mshibuya: Isn't it nice to being able to configure this per model basis?

app/views/rails_admin/main/index.html.haml
```diff
@@ -12,10 +12,22 @@
...
+    end
+    horiz_scroll_frozen_cols ||= 3 # by default, freeze checkboxes, links & first property (usually id?)
+    horiz_scroll_frozen_cols -= 1 unless checkboxes
+  else
+    horiz_scroll_frozen_cols = 0
+    sets = get_column_sets(properties)
+    properties = sets[params[:set].to_i] || []
+    other_left = ((params[:set].to_i - 1) >= 0) && sets[params[:set].to_i - 1].present?
+    other_right = sets[params[:set].to_i + 1].present?
+  end
```
mshibuya: Can these calculation logics be more simple?
If they needs to be that complex, consider moving it to outside of the view.

sbull:
Thanks for your feedback!

1. I had wanted to make this as low-impact as possible, so that the JS & CSS wouldn't even be loaded in the default configuration. If these are in separate files, I don't think it's possible to conditionally load them (assuming the right thing would be to make them part of the asset precompile pipeline)? But I can submit a change to do it the "right" way.

2. I expect that the most-used config will be global options, but I will look into making it work per-model as well.

3. I'll look into moving some of the calculation logic into the config class or helpers.

I hope to put some time into this in a few days. Thanks again!

sbull:
Hi @mshibuya, I've refactored this code to address your requests
1. Move JS & CSS into asset files.
2. Enable configuration at model level.
3. Move config logic out of view.

I've reworked the tests to account for these changes, and created new tests to check the model-level configuration. I've also rebased this PR. I can squash these commits if you'd like but I thought I'd leave them separate for now for you to review the changes from the original PR.

Let me know if other changes are needed. I don't really know how to fix the codeclimate "cognitive complexity" issue - perhaps split it up into multiple methods, but that seems like it could result in a fair amount of duplicate code.

Thanks!

sbull:
Hi @mshibuya , I was wondering if you could take a look at this, as it's been open for a while. Thanks!

mshibuya:
Thanks you for amazing work!
I'm planning to change the name of this feature to `sidescroll`.

twikah:
Hi @mshibuya, thanks for getting this merged! Are you planning on making a release soon? It would be great!

PR #3018, opened 2018-04-30
Fix dummy_app "rails db:seed" by replacing mlb-gem freebase dependency

sbull:


mshibuya:
Thank you for your work!

PR #3019, opened 2018-05-02
Add possibility to have custom search on model

RocKhalil:
You can now add a scope `rails_admin_search` that allows you to override the default search method. This will be useful for people who wants to search with joined model;

For example, if you're using `globalize` and the model's `name` column is translated, currently it fails to search for the correct one, but with `pg_search` and some minor basic querying with `pg_search` you'll be able to find your needed result.

Example of usage:
```ruby
  pg_search_scope :rails_admin_search,
    :against => [:id],
    :associated_against => {
      translations: :name
    },
    using: {
      tsearch: {any_word: true,},
      trigram: {threshold: 0.1}
    }
```

### DIFF
--- a/lib/rails_admin/adapters/active_record.rb
+++ b/lib/rails_admin/adapters/active_record.rb
@@ -32,11 +32,9 @@ module RailsAdmin
         scope = scope.where(primary_key => options[:bulk_ids]) if options[:bulk_ids]
 
         if options[:query]
-          begin
-            # if you're using pg_search, the method is not defined
-            # eventhough it's there
-            scope = scope.rails_admin_search(options[:query])
-          rescue
+          if self.config.search_scope
+            scope = scope.send(self.config.search_scope, options[:query])
+          else
             scope = query_scope(scope, options[:query])
           end
         end
--- a/lib/rails_admin/adapters/active_record.rb
+++ b/lib/rails_admin/adapters/active_record.rb
@@ -30,7 +30,17 @@ module RailsAdmin
         scope = scope.includes(options[:include]) if options[:include]
         scope = scope.limit(options[:limit]) if options[:limit]
         scope = scope.where(primary_key => options[:bulk_ids]) if options[:bulk_ids]
-        scope = query_scope(scope, options[:query]) if options[:query]
+
+        if options[:query]
+          begin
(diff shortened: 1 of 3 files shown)


lib/rails_admin/adapters/active_record.rb
```diff
@@ -30,7 +30,17 @@ def all(options = {}, scope = nil)
         scope = scope.includes(options[:include]) if options[:include]
         scope = scope.limit(options[:limit]) if options[:limit]
         scope = scope.where(primary_key => options[:bulk_ids]) if options[:bulk_ids]
-        scope = query_scope(scope, options[:query]) if options[:query]
+
+        if options[:query]
+          begin
+            # if you're using pg_search, the method is not defined
+            # eventhough it's there
+            scope = scope.rails_admin_search(options[:query])
```
mshibuya: The search method's name should be configurable via DSL.

RocKhalil: For the time being, it's good to have this one;
I'll work on making it configurable from Rails Admin's model config later on.

Let me know if you wish me to proceed with the second one before pulling.

-- Btw, this little fix saved me hours and hours of building a custom search page

mshibuya: Configurability is a key part of RailsAdmin, I don't merge this without it.
Also, having tests to make sure that this feature works is mandatory.

RocKhalil:
@mshibuya I've moved the `search_scope` to model's config;
However, I'm not sure how to write tests for it since I use Guard instead of RSpec;
Here's how it can be used:
```
config.model Player do
  search_scope :my_search
end

class Player < ApplicationRecord
  scope :my_search, -> (keyword) { where(name: keyword) }
end
```

mshibuya:
Based on your idea, 9abc7f929444049c1ac10b890e4aeb694e4c8e95 is now in the master.
Thanks!

RocKhalil:
@mshibuya excellent ! thanks :-)

mshibuya:
Ah, I forgot to mention that I've made slight modification to the DSL.
I've updated [the wiki](https://github.com/sferik/rails_admin/wiki/Custom-Search) accordingly.

Issue #2951, opened 2017-10-27
Compatibility issues with jQuery 3+ and jQuery UI prior to ver 12.0


jakephot:
Hello. I am getting this error on colsole when clicking on pulldown menu on RailsAdmin. The pulldown items are rendered but in the wrong place out of the default scope of the window.

```
Uncaught TypeError: elem.getClientRects is not a function
```

I believe this is due to incompatibility between jQuery 3+ and jQuery UI prior to version 12.0. RailsAdmin is dependent on `jquery-ui-rails (~> 5.0)`, and the latest 5.0.x is [bundled with `jQuery UI 1.11.4`](https://github.com/jquery-ui-rails/jquery-ui-rails/blob/master/VERSIONS.md). 

The details of this compatibility issue is [described here](https://github.com/vanderlee/colorpicker/issues/132).

LennonR:
+1 I'm getting this same error on Chrome

PR #3020, opened 2018-05-08
Bump jquery-ui-rails dependency to ~> 6.0

tinbka:
jquery-ui-rails starting at 6.0.0 fixes broken ui-autocomplete (drop-down inputs)
Related issue: https://github.com/sferik/rails_admin/issues/2951

### DIFF
--- a/rails_admin.gemspec
+++ b/rails_admin.gemspec
@@ -10,7 +10,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'font-awesome-rails', ['>= 3.0', '< 5']
   spec.add_dependency 'haml', '>= 4.0', '< 6'
   spec.add_dependency 'jquery-rails', ['>= 3.0', '< 5']
-  spec.add_dependency 'jquery-ui-rails', ['>= 5.0', '< 7']
+  spec.add_dependency 'jquery-ui-rails', ['>= 6.0', '< 7']
   spec.add_dependency 'kaminari', '>= 0.14', '< 2.0'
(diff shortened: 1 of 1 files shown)


mshibuya:
What does 'broken ui-autocomplete' mean?
RailsAdmin's ui-autocomplete used in filtering-select widget works fine with jquery-ui-rails 5.x.

tinbka:
@mshibuya which exact versions of `jquery-rails` and `jquery-ui-rails` you have rails_admin working fine with?

With 1.3.0 I had the error described in #2951 which leads to the filtering-select widget opening the dropdown list in a corner of the html page.

By googling a little it's turned out that the jquery version in jquery-rails gem which satisfies the dependency requirements of rails-admin 1.3.0 is now too high for the jquery-ui version which is shipped with jquery-ui-rails gem which satisfies them.

mshibuya:
I've checked with jquery-rails (4.3.3) and jquery-ui-rails (5.0.5).

Anyway, jquery-rails uses different asset names for different jQuery versions(`jquery` for jQuery 1.x, `jquery2` for jQuery 2.x, `jquery3` for jQuery 3.x), I think [RailsAdmin's reference of `jquery`](https://github.com/sferik/rails_admin/blob/master/app/assets/javascripts/rails_admin/rails_admin.js#L1) should always be resolved to jQuery 1.x.
jQuery 3.x won't be activated in RailsAdmin if you use jquery-rails properly.

Is there any chance that you have a jQuery 3.x file with the name of `jquery` ?

mshibuya:
Please reopen on update.

PR #3022, opened 2018-05-18
Fix for bulk delete shows error flash message on successful delete

aashish:
Hi,
This PR is a fix for the bulk delete shows error flash message on successful delete.


mshibuya:
Looks good, but tests are needed!

mshibuya:
After careful consideration, I think showing to users that some objects which they tried to delete did not be deleted is sometimes useful. Current implementation makes some sense.
Thanks for the PR, closing this.

PR #3027, opened 2018-05-29
Support filtering by multiple selected values

prusswan:
Currently the only way to filter by multiple selection (using `IN` operator) is to use enums, so I made a quick hack to get it to work with general integer columns. I can expand this further if there is more interest.


mshibuya:
Motivation is unclear, why you need this feature and how to use?

prusswan:
@mshibuya This is allow filtering using multiple values similar to `build_statement_for_enum`. Right now if multiple values are passed in for the same integer column, the filtering will work like "AND" (a == 8 AND a == 7). If the column is enum, then it works like "OR" (a == 8 OR a == 7)

mshibuya:
Change in `ra.filter-box.js` and `main_helper.rb` looks unrelated to the topic of PR, please separate them if you still need this feature.

PR #3028, opened 2018-05-31
use asset_path for codemirror asset links

mschiller:
Asset links without fingerprint are hard to handle in an normal rails production environment. So, hope, that this solution is good enough.


mshibuya:
Cherry-picked as f90e3a819d4b719badf620dca775938fa252f613.

PR #3029, opened 2018-06-06
Cancancan adapter is registered as cancan

glebtv:
Currently, cancancan cannot be set in ```config.authorize_with``` because it's registered with name ```cancan``` - same as ```cancan```

I am unsure why it's like this but it seems not to be intended.


glebtv:
Build failure seems to be unrelated to this..

PR #3041, opened 2018-07-05
code_reload

AnatolyShirykalov:
Fix reloading models when code changes. The problem is connected with pull request [#2999](https://github.com/sferik/rails_admin/pull/2999)
There the problem was solved partially. It fixed the "can't modify frozen array" exception on reload, but versions of models were still old.

### DIFF
--- a/lib/rails_admin/config.rb
+++ b/lib/rails_admin/config.rb
@@ -304,6 +304,14 @@ module RailsAdmin
         @registry.delete(key)
       end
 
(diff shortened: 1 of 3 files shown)


mshibuya:
Thanks for the fix, can you make CI pass?

AnatolyShirykalov:
@mshibuya done

mshibuya:
Great, thanks!

PR #3042, opened 2018-07-16
Use 1 as argument for count method instead of `:all`, as it is going to return the non …

absk1317:
…zero rows anyhow and reduce the overhead

### DIFF
--- a/lib/rails_admin/adapters/active_record.rb
+++ b/lib/rails_admin/adapters/active_record.rb
@@ -40,7 +40,7 @@ module RailsAdmin
       end
 
       def count(options = {}, scope = nil)
-        all(options.merge(limit: false, page: false), scope).count(:all)
+        all(options.merge(limit: false, page: false), scope).count(1)
       end
 
       def destroy(objects)


sfcgeorge:
This is a common misconception but COUNT(1) absolutely doesn't reduce overhead or go faster, if anything it's slower in some databases.

In this Rails code `:all` is an alias for `*`. And `count(*)` is optimised by the database as specified by the SQL standard; it's a special case—the * isn't actually an expression, it doesn't select all columns—so it's the fastest way to get a count. `count(1)` is not a special case so may or may not be optimised, for example Postgres has to check the value of `1` for every row to make sure it's not NULL (it never will be of course but it still checks).

This is a helpful breakdown of the syntax from [MS SQL Server docs](https://docs.microsoft.com/en-us/sql/t-sql/functions/count-transact-sql?view=sql-server-2017):

```sql
COUNT ( { [ [ ALL | DISTINCT ] expression ] | * } )
``` 

And the PostgreSQL docs list the two forms separately:

```sql
count(*) -- number of input rows
count(expression) -- number of input rows for which the value of expression is not null
```

I've seen benchmarks but for my own curiosity I tried a few queries in Postgres on some of our large tables: sure enough `count(*)` is slightly faster than `count(1)`.

In some databases `count(1)` is _also_ special cased and re-written into `count(*)` internally since so many people mistakenly do it. Oracle is one such database, but there's no guarantee; `count(*)` is the standard optimisation, `count(1)` may be a vendor specific optimisation.

**TLDR:** Always use `count(*)` as it's either faster than `count(1)` or identical.

https://docs.oracle.com/cd/E11882_01/server.112/e41084/functions039.htm#SQLRF00624
https://www.postgresql.org/docs/10/static/functions-aggregate.html
https://docs.microsoft.com/en-us/sql/t-sql/functions/count-transact-sql?view=sql-server-2017
http://sql.standout-dev.com/2017/01/count-vs-count1-again/#comment-119
https://stackoverflow.com/a/1221649/2311847
https://stackoverflow.com/a/50371522/2311847

sfcgeorge:
I expect you noticed the Rails Admin dashboard is slow, which is a good thing to notice and try to fix :)

Here is a really interesting article that explains why the count() function is slow [in many databases] and some alternatives such as using the query planner stats. But those stats are database specific so couldn't easily be done to speed up here. https://www.citusdata.com/blog/2016/10/12/count-performance/

PR #3049, opened 2018-07-30
Fix overriding session's option when using full-stack app.

### LINKED ISSUE #3048: Middleware stack mangled by 1.4.0
Updating to 1.4.0 broke my application due to [this commit](https://github.com/sferik/rails_admin/commit/5b55c756fdc9b660340176faf4f270b3eafbb86a) which installed middleware a standard Rails application already has, including cookie middleware which it then forcefully configures.

As a result, Rails now returns two session cookies in the HTTP response headers, and any cookie configuration you had previously gets thrown away.

In my case, this caused every app user to be logged out.

yskkin:
If developer is not using api mode, session's middleware configuration is up to a developer.
Otherwise it will overwrite configuration.

Fixes #3048

### DIFF
--- a/lib/rails_admin/engine.rb
+++ b/lib/rails_admin/engine.rb
@@ -26,7 +26,7 @@ module RailsAdmin
     initializer 'RailsAdmin setup middlewares' do |app|
       app.config.middleware.use ActionDispatch::Cookies
       app.config.middleware.use ActionDispatch::Flash
-      if app.config.api_only
+      if app.config.respond_to?(:api_only) && app.config.api_only
         app.config.session_store :cookie_store
         app.config.middleware.use ActionDispatch::Session::CookieStore, app.config.session_options
       end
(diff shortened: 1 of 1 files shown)


inopinatus:
I think all the middleware insertion should be configurable, because other than Rack::Pjax, all these are already present in a standard rails app and are therefore being duplicated, as can clearly be seen in the output of `rails middleware`. The overriding of cookie configuration is only the most egregious example.

Perhaps the middleware insertion belongs in the application's `config/initializers/rails_admin.rb` where it can be developer-managed, and not in the engine.rb where it is hard-coded.

wlipa:
It looks like ActionDispatch::Cookies and ActionDispatch::Flash would still be duplicated, as they are in Rails Admin 1.4 today.

mshibuya:
59478af9 will do the job, please check the master branch.

PR #3056, opened 2018-08-18
Fix export crash for models with JSON field

blaze182:
Hi everyone,

I experienced a crash when was exporting a model with a JSON field:

```ruby
NoMethodError:
undefined method `content_tag' for nil:NilClass'
# ./lib/rails_admin/config/fields/types/json.rb:17:in `block in
<class:Json>'
```

Failing lines:

```ruby
# lib/rails_admin/config/fields/types/json.rb
# ...
  register_instance_option :pretty_value do
    bindings[:view].content_tag(:pre) { formatted_value }.html_safe
  end
# ...
```

Apparently `bindings[:view]` is nil within export context.

Here's a proposed fix for that, I specified an `export_value` for json field.

Additionally, another test failed for model querying if JSON field is included, so I suppose it's more practical to return nil instead of raising JSON::ParserError in this case.

Thanks for an awesome product!

### DIFF
--- a/lib/rails_admin/config/fields/types/json.rb
+++ b/lib/rails_admin/config/fields/types/json.rb
@@ -17,6 +17,10 @@ module RailsAdmin
             bindings[:view].content_tag(:pre) { formatted_value }.html_safe
           end
 
+          register_instance_option :export_value do
+            formatted_value
+          end
+
           def parse_value(value)
             value.present? ? JSON.parse(value) : nil
           end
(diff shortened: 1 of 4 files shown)


mshibuya:
Thanks for the PR, fix for `export_value` looks good. Merged into master as 9fd8d03.

But for the change in `parse_input`, I don't think silently swallowing parse error is always useful for everybody, and it'a breaking change anyway so not merging in this time.

blaze182:
@mshibuya you can try adding json field to Player model and running specs, it will fail on existing query spec. Same happens in query user interface for any model containing json field, it unexpectedly crashes and needs an explicit `queryable: false` setting for json field. That’s why I consider it as a bug, probably you have better ideas to handle it (at least mentioning in docs)

PR #3059, opened 2018-08-27
Avoid using private API of PaperTrail

jaredbeck:
1. `version_class_name` is private API. Please use `paper_trail_options` instead.
2. Fixes deprecation warning re: PaperTrail.whodunnit

### DIFF
--- a/lib/rails_admin/extensions/paper_trail/auditing_adapter.rb
+++ b/lib/rails_admin/extensions/paper_trail/auditing_adapter.rb
@@ -37,6 +37,16 @@ module RailsAdmin
           created_at: :created_at,
           message: :event,
         }.freeze
+        E_VERSION_MODEL_NOT_SET = <<-EOS.strip_heredoc.freeze
+          Please set up PaperTrail's version model explicitly.
+
+              config.audit_with :paper_trail, 'User', 'PaperTrail::Version'
+
+          If you have configured a model to use a custom version class
+          (https://github.com/paper-trail-gem/paper_trail#6a-custom-version-classes)
+          that configuration will take precedence over what you specify in
+          `audit_with`.
+        EOS
 
         def self.setup
           raise('PaperTrail not found') unless defined?(::PaperTrail)
@@ -59,7 +69,7 @@ module RailsAdmin
           begin
             @version_class = version_class.to_s.constantize
           rescue NameError
-            raise "Please set up Papertrail's version model explicitly. Ex: config.audit_with :paper_trail, 'User', 'PaperTrail::Version'"
+            raise E_VERSION_MODEL_NOT_SET
           end
         end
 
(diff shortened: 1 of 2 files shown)


jaredbeck:
> version_class_name is private API. Please use paper_trail_options instead.

Everything in `paper_trail_options` is well-documented and is the main interface to PaperTrail, so is unlikely to change without warning.

`version_class_name` however is utterly private API never documented or intended for anyone else to use.

jaredbeck:
I also had some ideas about how to improve the test coverage for the PT adapter, but it'll require schema changes, and activerecord is happy on travis but mongoid is complaining and it's 0100 here and I just want to go to bed. 😴

jaredbeck:
Please restart job https://travis-ci.org/sferik/rails_admin/jobs/421174193 - suspected intermittent failure unrelated to this PR

mshibuya:
Great work, thanks!

rdunlop:
This change is causing issues for me, when I try to view the History of a model which does Not have papertrail, it fails with  
```
NoMethodError (undefined method `paper_trail_options' for #<Class:0x00007f80f6086500>
```

I've been able to resolve this by reverting this change.

jaredbeck:
Dang, sorry Robin. So rails_admin will use the `RailsAdmin::Extensions::PaperTrail::AuditingAdapter` even when you set `config.audit_with :something_else`?

rdunlop:
@jaredbeck I haven't had the opportunity to fully understand what's going on here, it could very well be a mis-configuration on my app...it's had many versions of rails_admin/pundit/papertrail over the years...and I just try to keep updating when possible...and then I run into breakages.

Here's what I have right now:
```
RailsAdmin.config do |config|
....
  config.authorize_with :pundit
  # If you want to track changes on your models:
  # config.audit_with :history, 'User'

  # Or with a PaperTrail: (you need to install it first)
  config.audit_with :paper_trail, 'User'
...
end
```

And then I fixed my current problem by reverting your fix:
```
# Fix issue where it tries to load paper_trail_options for a model which has none
# this reverts https://github.com/sferik/rails_admin/pull/3059/files
module RailsAdmin
  module Extensions
    module PaperTrail
      class AuditingAdapter
        def version_class_for(model)
          model_name = model.name
          klass = model_name.constantize.try(:version_class_name).try(:constantize)
          klass || @version_class
        end
      end
    end
  end
end

```

I also have a monkeypatch for rails_admin in order to change the pundit behavior, which could be involved (https://github.com/sudosu/rails_admin_pundit/issues/12#issuecomment-328357027)...but I don't think it should be. (I apologize that I haven't created a minimum reproduction scenario).

Note: the pertinent gems in my system right now are:
pundit (2.0.0)
rails_admin (1.4.2)
paper_trail (10.0.1)
(I do not have rails_admin_pundit or other paper_trail related gems installed).

I also tried it with paper_trail 9.2.0, same behavior.

PR #3066, opened 2018-09-12
Fix field is reseted to default if error happened

lazebny:
When error is raised for some object in `instance_eval(&value)` then `instance_variable_set("@#{option_name}_recurring", false)` won't be reset. As a result all other objects (from that model) will have a default value instead of custom value (even if they do not raise an error).


lazebny:
Sure thing, will do)

mshibuya:
Rubocop check is failing:
https://travis-ci.org/sferik/rails_admin/jobs/432086755#L1183

lazebny:
Sorry, I missed it. Checks are green now.

mshibuya:
Awesome, thanks!

PR #3081, opened 2018-11-01
Add Shrine support

akhsunna:
It would be great if rails_admin had also support of [shrine](https://github.com/shrinerb/shrine)


mshibuya:
Outstanding, thanks!

PR #3094, opened 2018-12-15
Fix most_recent_created order in dashboard. Fixes #3010

### LINKED ISSUE #3010: Wrong most_recent_created
now it is 
`@most_recent_created[t.model.name] = t.model.last.try(:created_at)`
rails 5 by default use `order by "model_name".id`. When using `uuid` as `id` it have wrong behaviour. I think this should use sorting order from model's settings
```
rails_admin do
    parent ParentModel
    list do
      sort_by :created_at
    end
end
  ...
```

dexion:


dexion:
I added new config param `:dashboard_sort_by` which can be setted like this
```
rails_admin do
    navigation_label 'Global Data'
    dashboard_sort_by :created_at
end
```
Now any user can  select preffered field for sorting (with or without indexes in DB). Users who use UUID may get right sorting order in their dashboards

Issue #2554, opened 2016-02-09
Cannot Query Mongoid/Mongo Hash via RailsAdmin


timchunght:
I currently have a model like the follow

`class Object
  include Mongoid::Document
  include Mongoid::Attributes::Dynamic

  field :data, :type => Hash
end
`
I would really like the ability to query the data field but Hash doesn't seem to be supported by rails_admin. Is there any fix or plugin in the community to solve this problem? Please advise. Thank you.

frontandstart:
Hi, I make this [PR](https://github.com/sferik/rails_admin/pull/3103).
I just edit hash as Json that RailsAdmin has

shanta31:
I'm facing the same issue. Is there a way to solve it ?

Issue - Not able to update mongoid hash field in rails admin (1.1.1)
Please advise. Thank you

PR #3103, opened 2019-01-17
Add support hash in RailsAdmin with Hash filed datatype

r3cha:
Hello, I need to use Hash with RailsAdmin and seems like can't edit hash filed and save it.
Its simple edit as plain Json
#2554 
[old stack overflow question](https://stackoverflow.com/questions/16783336/mongoid-admin-with-hash-and-array-field-types)
Here is how its look like:
screenshot 2019-01-17 at 14 00 47
screenshot 2019-01-17 at 14 00 54
I have no idea broke it or not but it's work for me.

### DIFF
--- a/lib/rails_admin/adapters/mongoid/property.rb
+++ b/lib/rails_admin/adapters/mongoid/property.rb
@@ -19,7 +19,6 @@ module RailsAdmin
         end
 
         def type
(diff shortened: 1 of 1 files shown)


frontandstart:
Add suppler of Hash

PR #3112, opened 2019-02-27
improve readme

cprodhomme:
The goal is to make it easier to find the demo and documentation

That will be great if you add some `tags` and the url of the demo in the header of the repositery.

Like activeadmin : 
capture d ecran 2019-02-27 a 12 10 44


mshibuya:
Thanks for contribution!

@sferik Can you please take care of the tags?

Issue #2099, opened 2014-10-24
Problem with blanks and nils with unique indexes


justin808:
rails_admin is creating new records with "" (empty string) rather than nil, and for a column where there is a unique index, allowing for nil, then this leads to a non-unique index violation, as the second record with a "" is reported as not unique.

wallace:
:+1:

PR #3121, opened 2019-04-02
Does not parse empty enum value

rodolfoliviero:
Prevent parse empty enum value that cause to automatic fill enum field with blue instead of nil.

```ruby
  enum color: {
    blue: 0,
    red: 1
  }
```


sedubois:
CI reports `lib/rails_admin/config/fields/types/active_record_enum.rb:46:1: C: Trailing whitespace detected.`

mshibuya:
Merging into #2099.

mshibuya:
This should be no longer an issue since we are supporting only Rails >= 5.0 now.

PR #3123, opened 2019-04-12
Adding id locale translation

yogahp:
I need to add id locale translation to rails_admin. So i create this PR. I've add id locale translation plus the test


mshibuya:
Thanks for PR, but we maintain all translations in wiki:
Please add yours there.

PR #3127, opened 2019-05-03
Add Bootstrap 4 compatibility

DerekK19:
Added function to convert px units to em (divide by 16). Modified any styles that fail to precompile with Bootstrap 4 due to mixing px and em units - These now call the px_to_em function to convert the px units to em.

Note: I have not converted every px unit to em, only those that caused a rake assets:precompile failure. Also note there are no specs for this change - I kept running rake assets:precompile until it succeeded, changing one px unit at a time.

### DIFF
--- a/app/assets/stylesheets/rails_admin/base/theming.scss
+++ b/app/assets/stylesheets/rails_admin/base/theming.scss
@@ -3,7 +3,7 @@
   Keep it clean, people
 */
 
-$avatar-size: 30px;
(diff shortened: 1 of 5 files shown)


DerekK19:
I'm closing this, because I am now finding that it doesn't work. I'm now seeing smaller text than I'd expect. I'd appreciate any comments anyone else might have about integrating Bootstrap 4 with Rail Admin. It seems it's harder than I imagined

PR #3128, opened 2019-05-03
Bootstrap 4

DerekK19:
To get text the correct size on the web page, set the html font-size to 1em - all text on the page is relative to this size. Don't ceil or floor font sizes - you'd do this for px units, but for rem units we want the decimals. Form input boxes add 2 (pixels) to the line height - Need to convert this 2 to ems. The Admin pages use the bootstrap folder directly (apps use rails_admin), so need to include the functions.scss file in bootstrap. the px_to_em function returns rem units not em siren rem units are used elsewhere and we'd get more "bootstrap can't combine units" errors


mshibuya:
You're saying that your rails app(using RailsAdmin for admin backend) breaks when used with Bootstrap 4, assets failing to compile or showing wrong style - is that correct?
How can I reproduce the problem you're experiencing?

DerekK19:
Bootstrap 4 doesn't allow measurement units to be mixed (you can define the width of, say, a text in terms of pixels or of ems or of rems, but you can't define one text in terms of pixels and then define another text in terms of the first text plus a number of ems. Or you can't say that a text is 3 ems high then say a button is the same height as the text plus 2 pixels - you always have to use the same units.

In Rails Admin, there is a file with a number of variables (variables.scss) that defines the sizes of text, and so on. This file needs to consistently use the same units. Rather than go through the file and change every pixel measurements to the corresponding number of rems, I added a new function that converts a number of pixels to the same number of rems (16 pixels = 1 rem) and I applied that function to all the pixel measurements. Now all sizes are defined in terms of rems. Also note that rems are often small decimal numbers (while pixels are almost always integers) so flooring and ceiling rems is not a good thing to do.

I found that just changing from Bootstrap 3 to Bootstrap 4 in my GemFile and then asset precompiling threw up a large number of errors all caused by mixed measurement units. I changed definitions one at a time until I resolved all the errors.

DerekK19:
I should also say that my site (https://library.gordonknight.co.uk) uses Rails Admin, but only for administrators. Regular users see pages that are defined as rails views.

mshibuya:
Hmm... I could not reproduce.

My attempt went like this:
1. Using [dummy_app](https://github.com/sferik/rails_admin/tree/master/spec/dummy_app), added `gem 'bootstrap', '~> 4.3.1'` to Gemfile
2. Renamed `app/assets/stylesheets/application.css` to `application.scss` and added `@import "bootstrap";` there.
3. Run `rake assets:precompile`, nothing broke and `application-<hash>.css` is generated containing styles from Bootstrap 4.

What's the difference with yours?

DerekK19:
I'm going to close this pull request.

I'm now very confused. I have just reverted my gemfile back to using the standard rails admin gem (now at version 1.4.2). I clobbered and cleaned my assets and then pre-compiled and my assets precompile successfully. All I did was change my gemfile.

I can only think that I must have changed something else in the past few weeks - probably after I modified rails-admin and that negated the need for my rails-admin changes. It must be something in my own css files that was causing my problems. Going back over my changes, I can't see anything obvious that I have changed.

I'm very sorry for wasting your time on this.

PR #3135, opened 2019-05-11
Fix localization of "time ago in words"

sedubois:
Not all languages have the same order as English ("x months ago" <-> "il y a x mois")


mshibuya:
It's giving SyntaxError...

sedubois:
should be fixed now

mshibuya:
Looks good, thanks!

Issue #2960, opened 2017-11-22
Add rel="noopener" to all inks that have target="_blank"


tlkiong:
`rel="noopener"` should be added to links containing `target="_blank"` as a precaution against `reverse tabnabbing`. For more information, please refer to the following article:
["Target="_blank" - the most underestimated vulnerability ever"](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/)

For example:
HTML: `<a href="http://example.com" target="_blank">Link</a>`
Fix: `<a href="http://example.com" target="_blank" rel="noopene">Link</a>`

On a side note, `noreferrer` issue has been fixed in Firefox 52.

I suggest to not cater to older browsers as it is futile because there are numerous other security exploits those browsers are susceptible to. If the user wants to be secure, then the user should be using the latest browser.

---

For further reading:
1. ["The performance benefits of rel=noopener"](https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/)
2. ["About rel=noopener - What problems does it solve?"](https://mathiasbynens.github.io/rel-noopener/)

mshibuya:
Fixed by #3169.

PR #3169, opened 2019-07-14
Add rel='noopener noreferrer' to links with target=_blank

olleolleolle:
This PR adds security-related HTML properties to links with the `target=_blank` attribute.

See #2960 for more details.

<details>

  - Cater to old Firefox, before 52, with rel=noreferrer as well
    https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/52#HTML

</details>


mshibuya:
Thanks!

PR #3180, opened 2019-08-05
Add missing "en.admin.misc.ago"

mikker:
Used in https://github.com/sferik/rails_admin/blob/master/app/views/rails_admin/main/dashboard.html.haml#L21


mshibuya:
Thanks!

PR #3188, opened 2019-08-21
Bump haml version requirement in gemspec

fooki:
haml versions below 5.1.1 will crash with rails 6.0.0 due to a missing Erubi constant. 

See: https://github.com/haml/haml/issues/1011

background: rails_admin crashed for me because i had an old (but still accepted according to gemspec) haml version laying around when installing.


mshibuya:
No, this does not mean that RalisAdmin is not compatible with haml < 5.1.1.
Library dependency should be broad as much as possible, and if users encounter a problem they can apply their own strategy to avoid versions with the problem.

PR #3189, opened 2019-09-01
Allow hide History in dashboard

elsupergomez:


mshibuya:
Could you add some tests for this?

mshibuya:
Thanks!

PR #3207, opened 2019-10-15
Make vendorized FontAwesome to work with Sprockets 4, fix failing tests

bodrovis:
Resolves #3204

1. With Sprockets 4 `font-awesome.css.erb` file is unreadable. Converting it to `.scss` fixes the issues (huge thanks to @cselmer for help)
2. Fixed failing tests on Windows (some gems were not installed on x64 machines, also TZ data should be included)
3. Fixed failing tests for CanCan (kept getting `CanCan is undefined at ./spec/integration/authorization/cancancan_spec.rb:5:in class:Ability'`)
4. Created manifest file so that dummy app plays nicely with Sprockets 4

PR #3210, opened 2019-10-17
Fix the case where username is nil, not whodunnit

okuramasafumi:
I encounter the problem where paper_trails's whodunnit isn't displayed.
It turned out that `username` method returns nil.
`try` method returns nil when there's no method, so when user doesn't have `email` method,
`@user_class.find(@version.whodunnit).try(:email)` returns nil.
Then, try this out:
`nil rescue nil || :foo`
This returns nil, surprisingly.
This means if user doesn't have email method, username is always nil.
So my fix is like below:
`(nil rescue nil) || :foo`
This returns `:foo` as expected.

### DIFF
--- a/lib/rails_admin/extensions/paper_trail/auditing_adapter.rb
+++ b/lib/rails_admin/extensions/paper_trail/auditing_adapter.rb
@@ -23,7 +23,7 @@ module RailsAdmin
         end
 
         def username
-          @user_class.find(@version.whodunnit).try(:email) rescue nil || @version.whodunnit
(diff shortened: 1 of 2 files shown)


mshibuya:
Could you write a spec for this issue?

okuramasafumi:
@mshibuya I'd like to add tests but CI seems to fail. Could you fix it?

mshibuya:
It's already fixed in the master.
#3207

okuramasafumi:
@mshibuya Great, thanks!

mshibuya:
Thanks!

PR #3213, opened 2019-10-28
Only hide empty values if compact mode is enabled

fuegas:
If a value is an empty string, it currently is hidden regardless of the setting `compact_show_view`. 

I think an empty string should only be hidden if compact mode is enabled. This PR enables that behaviour.


mshibuya:
Could you write a spec for this issue? Best place will be:
https://github.com/sferik/rails_admin/blob/master/spec/integration/actions/show_spec.rb

fuegas:
@mshibuya : Tests have been added. When you run these without the parentheses in the view the test `shows blank fields` will fail:
```
  2) Show action when compact_show_view is disabled shows blank fields
     Failure/Error: is_expected.to have_css('.logo_url_field')
       expected #has_css?(".logo_url_field") to return true, got false
     # ./spec/integration/actions/show_spec.rb:95:in `block (3 levels) in <top (required)>'
```

mshibuya:
Perfect, thanks!

PR #3217, opened 2019-11-06
Avoid mutable constant warning in a test

olleolleolle:
This PR avoids an explicitly added exemption from a RuboCop warning.


mshibuya:
👍

PR #3221, opened 2019-11-14
Add class '.table-hover' to tables

nard-tech:
I added class `.table-striped` to tables for visibility of table.


mshibuya:
Thanks!

PR #3227, opened 2019-11-28
Use kaminary param_name for pagination

elsupergomez:
Kaminary has a configuration to custom `param_name`, and this change use that configuration like does [Main index](https://github.com/sferik/rails_admin/blob/9abc1be0c1fee006a7a746583461891911fe9330/app/controllers/rails_admin/main_controller.rb#L125) instead of use hard code `:page` name


mshibuya:
Thanks, but we need test code for this!

mshibuya:
Thanks!

PR #3232, opened 2019-12-05
Unify variable of #method_missing and #respond_to?

nard-tech:
Unify first variable of `#method_missing`  and `#respond_to?` to `method_name`.


mshibuya:
👍

PR #3238, opened 2020-01-02
Case-insensitive date parsing when delocalising

tom-lord:
The recent `v2.0.1` release of `rails_admin` included [updating the `momentjs` library](https://github.com/sferik/rails_admin/pull/3201).

A non-obvious side-affect of this is that the newest version of `momentjs` localises dates differently - for example, Spanish days and months are now (correctly?) [formatted as lower-case](https://github.com/moment/moment/blob/develop/locale/es.js) instead of upper-case.

(E.g. Rails admin `2.0.0` formatted today as "2 de Enero de 2020", but version `2.0.1` formats today as "2 de enero de 2020"). (But I'm using `fr` locale in this spec for [convenience](https://github.com/sferik/rails_admin/tree/master/spec/dummy_app/config/locales), since the tests already define it.)

However, `rails-admin` assumed an _exact match_ between the ruby-translations and javascript-translations. Any project which formats those dates differently (i.e. by capitalising them) no longer works with rails-admin.

Ideally, I feel the "right" solution here would be to **not send the date as a human-formatted-string** to the server (!!!) and instead use some well-defined ISO 8601 format; thus eliminating this whole complication.

But for now - as an immediate fix - I propose just making this substitution case insensitive.

---

N.B. The code climate "failure" doesn't actually change the existing code structure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 [Full Changelog](https://github.com/sferik/rails_admin/compare/v2.0.1...HEAD)
 
+### Fixed
+- Improved I18n date format handling ([#3238](https://github.com/sferik/rails_admin/pull/3238))
+
 
 ## [2.0.1](https://github.com/sferik/rails_admin/tree/v2.0.1) - 2019-12-31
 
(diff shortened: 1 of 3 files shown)


mshibuya:
Closing with the same reason as https://github.com/sferik/rails_admin/pull/3294#issuecomment-787383687.

PR #3246, opened 2020-01-29
Update readme

VSPPedro:
Hi, I noticed that rails_admin supports Ruby 2.7 and this information was not in the readme.

:smile: 

Thanks for this awesome gem! :+1:


mshibuya:
Thanks 👍

PR #3249, opened 2020-02-10
Remove not using of yell_for_non_accessible_fields

soruma:
`yell_for_non_accessible_fields` is no longer needed by #1622.
I think this configure is delete.


mshibuya:
Thanks!

PR #3255, opened 2020-02-19
Active storage variants in Rails 6>

kallelundgren93:
Since upgrading to Rails 6, my team changed to use variants in active_storage with vips. In vips the old:

```resize: "100x100"```

results in a broken file for rails_admin users that uses vips.
Instead I've added a check to see if the project is using ruby-vips and in that case they will be using the resize method:

```resize_to_limit: [100, 100]```

Another way to go to check it, is to check:
```config.active_storage.variant_processor ```
if that is preferrable.


mshibuya:
Merged in, thanks!

PR #3257, opened 2020-02-24
Upgrading to Shrine 3.x

akhsunna:
Considering that `derivatives` plugin replaced `versions` plugin in **shrine 3.x**, some code should be updated.

-------------------------------------------------

Also, I added `link_name` option to `RailsAdmin::Config::Fields::Types::FileUpload`. I hope it's okay. I did it because currently links for shrine fields don't look nice:

image

### DIFF
--- a/spec/dummy_app/app/active_record/shrine_versioning_uploader.rb
+++ b/spec/dummy_app/app/active_record/shrine_versioning_uploader.rb
@@ -2,17 +2,14 @@ class ShrineVersioningUploader < Shrine
   plugin :activerecord
 
   plugin :cached_attachment_data
-  plugin :delete_raw
   plugin :determine_mime_type
   plugin :pretty_location
-  plugin :processing
   plugin :remove_attachment
-  plugin :versions
+  plugin :derivatives
 
-  process(:store) do |io|
+  Attacher.derivatives_processor do |original|
     {
-      original: io,
-      thumb: FakeIO.another_version(io, :thumb),
+      thumb: FakeIO.new('', filename: File.basename(original.path), content_type: File.extname(original.path)),
     }
   end
 end
--- a/spec/dummy_app/app/active_record/shrine_versioning_uploader.rb
+++ b/spec/dummy_app/app/active_record/shrine_versioning_uploader.rb
@@ -5,11 +5,14 @@ class ShrineVersioningUploader < Shrine
   plugin :determine_mime_type
   plugin :pretty_location
   plugin :remove_attachment
-  plugin :derivatives
 
-  Attacher.derivatives_processor do |original|
-    {
-      thumb: FakeIO.new('', filename: File.basename(original.path), content_type: File.extname(original.path)),
-    }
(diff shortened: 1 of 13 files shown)


prem-prakash:
tested here, did not work

when there is already an attached file, the attachment data is not loaded on a hidden field, and if the user did upload a new file, it gives an error:

```
Shrine::InvalidFile ("" is not a valid IO object (it doesn't respond to #read, #eof?, #rewind, #close)):
```

prem-prakash:
I was able to make it work.
It seems that the option `cache: true` is necessary for the plugin model

```ruby
  plugin :model, cache: true # false won't work
```

akhsunna:
Hi @prem-prakash , thanks for your response

But I cannot reproduce it, what are the steps?

1. `gem 'rails_admin', '~> 2.0'`
2. upload the file
3. `gem 'rails_admin', github: 'LinkUpStudioUA/rails_admin', branch: 'support_shrine_v3'`
4. open the edit page again

Everything works for me, even if I use `plugin :model, cache: false`

akhsunna:
So I fixed the `cache_method` and tests for ruby 2.2.

**Travis CI**: I'm not sure why the tests randomly fail on the "nested form" part :(
**Code Climate**: this is how it was before, so probably we can ignore these issues within this pull request

spec/dummy_app/app/active_record/shrine_versioning_uploader.rb
```diff
@@ -2,17 +2,17 @@ class ShrineVersioningUploader < Shrine
...
-  plugin :processing
   plugin :remove_attachment
-  plugin :versions
 
-  process(:store) do |io|
-    {
-      original: io,
-      thumb: FakeIO.another_version(io, :thumb),
-    }
+  if Gem.loaded_specs['shrine'].version >= Gem::Version.create('3')
```
prem-prakash, line 9: shouldn't there be an `else` with the versions plugin?

akhsunna: `thumb` method will work only with derivatives
https://github.com/sferik/rails_admin/blob/e4d970e96a845f20ed898ebd6456062bb13a422c/lib/rails_admin/config/fields/types/shrine.rb#L10-L13

It won't work with version of shrine older than v3. So I just skipped tests about image versions:
https://github.com/sferik/rails_admin/blob/e4d970e96a845f20ed898ebd6456062bb13a422c/spec/rails_admin/config/fields/types/shrine_spec.rb#L14-L20

@prem-prakash, do you think we need to support thumbs for older versions of shrine too?

prem-prakash: @akhsunna I would say that it would be a good practice at least to not break with older versions, what do you think?

akhsunna: Okay, maybe it's worth it :) Will add it soon

mshibuya:
Merged in, since this is necessary in attempt to run Ruby 3.0 CI build.
Thanks for the contribution!

PR #3259, opened 2020-02-26
Avoid using jquery3 (some JS errors break the UX)

chalofa:


mshibuya:
`jquery-rails` provides [1.x version of JQuery](https://github.com/rails/jquery-rails/blob/master/vendor/assets/javascripts/jquery.js#L2) by referring to just `jquery`. This should just be your app have 3.x JQuery as `jquery`.

PR #3262, opened 2020-03-18
Fix polymorphic association field with namespaced models

dalpo:
The current code is generating invalid jQuery selectors for models with
a namespace.

For instance a model `Users::Account` will generate a `#users::account-js-options` DOM ID into the polymorphic association field template, causing an invalid jQuery selectors and breaking the related javascript code.


mshibuya:
Closing in favor of #3377.

PR #3266, opened 2020-03-25
Fix for authorization issues with pundit

ceberhardt:
- authorized? in _submit_buttons.html.haml called incorrectly:
  "edit" is a member action, therefore "authorized?" has to be called including @object, otherwise an error is thrown upon authorization in policy if @record is used
- Fix authorization_key for bulk_delete, was :destroy which resulted in error upon authorization, collection != member


ceberhardt:
I am switching to CanCanCan, Pundit support seem to be limited.

Thanks to all past contributors!

PR #3269, opened 2020-04-03
Fix to paper_trail anditing_adapter for custom names for the versions…

StudizDK:
… and version associations


mshibuya:
Closing in favor of #3354.

PR #3294, opened 2020-07-20
Add boundaries around date translations matching

danielveleba:
eg. June in Czech is substring of July (Červen - Červenec), so w/o the boundaries  `Červenec` translated to `Juneec`, which raised exception in TZ processing and resulted in `nil` value.

Sorry, don't have time to write tests.


mshibuya:
Sadly this will lead to another issue, since in some languages date representation are not separated by delimiters(e.g. `2021年2月28日` in Japanese).
Let me merge this into an attempt planned in #3344.

PR #3295, opened 2020-08-02
Show Correct Filename For Multiple Attachments

RongRongTeng:
## Show Correct File Name For Multiple Attachments

Now multiple attachments of ActiveStroage with show `ActiveStroage object` when the attachment is not an image.
Fix to show it with the `filename`.


mshibuya:
Looks good, thanks!

PR #3300, opened 2020-08-20
Fix for enums where keys and values differ

coorasse:
If the enums are defined with different values on the database, the current implementation doesn't show the value selected when editing a field.
This patch includes a regression by changing the existing keys for the enum field


prem-prakash:
@coorasse build is broken

PR #3309, opened 2020-09-30
Use RailsAdmin.I18n for multiselect translations

zaknafain:
**Problem:**
I tried to translate a custom action into German but had problems translating the multi select.
I found out that the English words were hard coded.
There was also a placeholder in the search form, that was reading undefined.

**Solution:**
I replaced the hard coded English words with the already used RailsAdmin.I18n.t function and moved them to the en.yml.
I added a translation for the undefined search form placeholder.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,10 +75,10 @@ matrix:
     - rvm: ruby-head
       env: CI_ORM=active_record CI_DB_ADAPTER=sqlite3
       gemfile: gemfiles/rails_6.0.gemfile
-    - rvm: jruby-9.2.11.0
+    - rvm: jruby-9.2.13.0
       env: CI_ORM=mongoid
       gemfile: gemfiles/rails_6.0.gemfile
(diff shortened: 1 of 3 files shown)


zaknafain:
Sadly there is an `Error in `/usr/lib/jvm/java-8-oracle/bin/java': double free or corruption (out)` error for jruby.
I never worked with jruby but I am pretty sure, that this is not a result of my changes.

zaknafain:
I had to update the version of jruby to fix the error mentioned above.

jubilee2:
> I had to update the version of jruby to fix the error mentioned above.
> This would close #3311

thank you! i am out

mshibuya:
Translations are already passed to JS [here](https://github.com/sferik/rails_admin/blob/master/app/views/rails_admin/main/_form_filtering_multiselect.html.haml#L31-L38), though it does not include `add` and `remove` keys...
Please try with community-provided translations.

PR #3311, opened 2020-10-01
CI: Update to JRuby 9.2.13.0

jubilee2:
fix double free or corruption (out) error in JRuby


mshibuya:
Adding sleeps will not help, so merged only 9b8eed5 into master. Thanks!

PR #3315, opened 2020-10-22
Multi select Search Placeholder Fix

zaknafain:
While trying to translate the multi select, I stumbled over an `undefined` search placeholder for the input field.
I found out it was a combination of a missing translation option in a plugin and a missing default option text in rails_admin.

This is why I added the missing default translation option for the multi select js.


mshibuya:
Makes sense, thanks!

PR #3318, opened 2020-11-11
Enhancement: Default filter operator for fields

mitchkettzend:
**Issue**
When a filter is initially added on the filter criteria on a list view, the default operator is `nil` or empty which makes it necessary for the user to select an operator each time they add a filter.  This can be a bit cumbersome especially if users search a field by a particular operator the majority of the time.

**Proposed change**
A field option to set a default filter operator for a field which populates the filter operator select box when initially added to the filter criteria.  If one is not specified then it will default to the original behavior of setting the operator to `nil` or empty.


mshibuya:
Looks good, thanks!

PR #3340, opened 2021-02-19
Hide present / blank filter options for required fields

codealchemy:
### Issue

For cases where a field is required (via validations, model config, or otherwise) and filterable, the associated filter dropdown includes `present` and `blank` options that are unnecessary as the field is always expected to be present.

### Proposed change

Pass the `required` field from the model through to the filter box javascript, keeping the existing order but only appending the `present` and `blank` options when the field is not required.

📝 I'm not sure how to best test the `js` changes aside from the tests added to `index_spec.rb` (which put it over the `rubocop` TODO limit, so I added that file as an exception instead of bumping the max any further).

---

Example using the spec `dummy_app`

#### Required field (`name`)
Screen Shot 2021-02-18 at 15 34 33

#### Optional field (`team`)
Screen Shot 2021-02-18 at 15 34 47

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -131,6 +131,8 @@ Lint/UnneededSplatExpansion:
 # ExcludedMethods: refine
 Metrics/BlockLength:
   Max: 1081
+  Exclude:
+    - 'spec/integration/actions/index_spec.rb'
 
 # Offense count: 7
 Naming/AccessorMethodName:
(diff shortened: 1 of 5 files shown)


codealchemy:
Failures appear unrelated to this change (and are happening on [my fork](https://github.com/codealchemy/rails_admin/actions/runs/579768762) as well).

mshibuya:
Nice, thanks!

PR #3343, opened 2021-02-22
Update database cleaner to >= 2.0

codealchemy:
Followup to https://github.com/sferik/rails_admin/pull/3341

[Related changes to `database_cleaner` gem](https://github.com/DatabaseCleaner/database_cleaner/blob/master/History.rdoc#200beta-2020-04-05-)

`database_cleaner` was split into separate adapter gems - the two needed for this project are:
- [`database_cleaner-active_record`](https://github.com/DatabaseCleaner/database_cleaner-active_record)
- [`database_cleaner-mongoid`](https://github.com/DatabaseCleaner/database_cleaner-mongoid)

This also removes Ruby 2.2 from the CI build, since `database_cleaner` uses the safe navigation operator which was introduced in Ruby 2.3 (see [related CI failure here](https://github.com/codealchemy/rails_admin/runs/1949121193?check_suite_focus=true)). Maintaining a Ruby 2.2 build has caused issues elsewhere as well (see related discussion in https://github.com/sferik/rails_admin/pull/3256#issuecomment-608205133), and 2.2 itself is no longer maintained as of [March, 2018](https://www.ruby-lang.org/en/downloads/branches/).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,7 +22,8 @@ end
 group :test do
   gem 'cancancan', '~> 3.0'
   gem 'carrierwave', ['>= 2.0.0.rc', '< 3']
-  gem 'database_cleaner', ['>= 1.2', '!= 1.4.0', '!= 1.5.0', '< 2.0']
+  gem 'database_cleaner-mongoid', '>= 2.0', require: false
+  gem 'database_cleaner-active_record', '>= 2.0', require: false
   gem 'dragonfly', '~> 1.0'
   gem 'factory_bot', '>= 4.2'
   gem 'generator_spec', '>= 0.8'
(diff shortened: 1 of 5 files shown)


mshibuya:
Thank you, but dropping Ruby support can only be done in major releases. I've set the 3.0.0 milestone.

codealchemy:
@mshibuya 👍🏼 understood. Is there a different branch this should target in the meantime? Also would the `3.0.0` milestone include removing support for libraries that are no longer maintained (ex. [`Paperclip`](https://github.com/thoughtbot/paperclip#deprecated))

mshibuya:
I've made the master branch 3.0.0 ready. Could you rebase this to that?

mshibuya:
Thanks ❤️

PR #3354, opened 2021-04-22
Support alternative paper_trail versions association name

fuegas:
When using PaperTrail as an auditing gem, you can specify an alternative versions association by defining the `versions` hash in the `has_paper_trail` call. Initialization of PaperTrail (per model) stores the name of the association (since https://github.com/paper-trail-gem/paper_trail/commit/3032b53, 23rd of July 2011) in `versions_association_name`. Using this method we can retrieve the history of a model when it uses a different name for the versions association.

For example, using the following configuration won't break the auditing_adapter with this change:
```ruby
has_paper_trail(
  version: :paper_trail_version,
  versions: {
    name: :paper_trail_versions,
  },
)
```

### DIFF
--- a/lib/rails_admin/extensions/paper_trail/auditing_adapter.rb
+++ b/lib/rails_admin/extensions/paper_trail/auditing_adapter.rb
@@ -116,7 +116,7 @@ module RailsAdmin
 
           current_page = page.presence || '1'
 
-          versions = object.nil? ? versions_for_model(model) : object.versions
+          versions = object.nil? ? versions_for_model(model) : object.public_send(model.model.versions_association_name)
           versions = versions.where('event LIKE ?', "%#{query}%") if query.present?
           versions = versions.order(sort_reverse == 'true' ? "#{sort} DESC" : sort)
           versions = all ? versions : versions.send(Kaminari.config.page_method_name, current_page).per(per_page)


lib/rails_admin/extensions/paper_trail/auditing_adapter.rb
```diff
@@ -116,7 +116,7 @@ def listing_for_model_or_object(model, object, query, sort, sort_reverse, all, p
 
           current_page = page.presence || '1'
 
-          versions = object.nil? ? versions_for_model(model) : object.versions
+          versions = object.nil? ? versions_for_model(model) : object.send(model.model.versions_association_name)
```
codealchemy, line 119: I believe `.versions_association_name` is public

```suggestion
          versions = object.nil? ? versions_for_model(model) : object.public_send(model.model.versions_association_name)
```

Also is this covered by any existing tests?

codealchemy: Sidenote - this solves the same problem as https://github.com/sferik/rails_admin/pull/3269 with a _slightly_ different approach (IIRC `object.version_association_name` will execute a SQL query, while calling from the model does not)

fuegas: It is public indeed, I'll change it to public_send. I used send to keep it in line with the other `send` usage in the file.

There was no test covering the versions usage, only one regarding the user used for the audit.

fuegas:
@codealchemy would you consider either of proposed solutions? 🙂

codealchemy:
> @codealchemy would you consider either of proposed solutions? 🙂

I'm afraid anything beyond providing feedback here is out of my hands cc: @mshibuya

fuegas:
Hi @mshibuya, any chance you might have some time to look at this (small) change? 😄

mshibuya:
Thanks!

PR #3365, opened 2021-06-18
Show "no records found" when a list is empty.

matthewfu:
This is a minor view change for list.

When there is no record found in list action, show no records found to give user consistent UI result with their request.

original issue: https://github.com/sferik/rails_admin/issues/3329


mshibuya:
Thanks!

PR #3370, opened 2021-07-15
fixed jquery3 loading issue

ruturajdesai92:
Failing to load specific jQuery version because of missing quotes.


mshibuya:
Thanks!

PR #3372, opened 2021-08-03
Fix missing select options for enum filters

codealchemy:
Following https://github.com/sferik/rails_admin/commit/555f7783f2255ddc736141c410c5bdaee074887a, the (single) select dropdown for `enum`s no longer includes the available values (though multiselect still works)

This ensures we append the available enum options to the single select dropdown.

### Before

Screen Shot 2021-08-03 at 13 25 59

### After

Screen Shot 2021-08-03 at 13 26 14


mshibuya:
Sorry for this issue, thanks for fixing!

PR #3377, opened 2021-08-11
Fix #3376: Polymorphic associations don't work with namespaced classes

### LINKED ISSUE #3376: Polymorphic associations don't work with namespaced classes
Given a set of models:

```ruby
class Picture < ApplicationRecord
  belongs_to :imageable, polymorphic: true, inverse_of: :pictures
end

module Foo
  class Employee < ApplicationRecord
    has_many :pictures, as: :imageable, inverse_of: :imageable
  end  
end

module Bar
  class Product < ApplicationRecord
    has_many :pictures, as: :imageable, inverse_of: :imageable
  end
end
```

When creating a new `Picture`, in the form, after selecting the type, RailsAdmin will fail to load the records for the associations. The problem is that RailsAdmin will get the name of the model:

https://github.com/sferik/rails_admin/blob/238f18ee2386f9858670b7995dcb628b8fe6bde9/lib/rails_admin/config/fields/types/polymorphic_association.rb#L66

Which is then used as the `id` of an HTML tag:

https://github.com/sferik/rails_admin/blob/238f18ee2386f9858670b7995dcb628b8fe6bde9/app/views/rails_admin/main/_form_polymorphic_association.html.haml#L2

https://github.com/sferik/rails_admin/blob/238f18ee2386f9858670b7995dcb628b8fe6bde9/app/views/rails_admin/main/_form_polymorphic_association.html.haml#L12-L15

https://github.com/sferik/rails_admin/blob/238f18ee2386f9858670b7995dcb628b8fe6bde9/app/views/rails_admin/main/_form_polymorphic_association.html.haml#L23-L24

When calling `.name` on a model declared in a module, the result will contain colons, i.e. `Bar::Product`. A string containing `:` is not a valid to fetch an HTML tag using the `id` attribute, as the JavaScript does here:

https://github.com/sferik/rails_admin/blob/238f18ee2386f9858670b7995dcb628b8fe6bde9/app/assets/javascripts/rails_admin/ra.widgets.js#L205-L206

I'm using:
rails_admin: 2.2.1
rails: 6.1.4
ruby: 3.0.0

jacob-carlborg-apoex:


mshibuya:
To make mongoid builds pass, you need to add the models for mongoid as well.

jacob-carlborg-apoex:
I'll give that a try.

mshibuya:
Amazing job, thanks!

jacob-carlborg-apoex:
Thanks.

PR #3378, opened 2021-08-11
Add HTML5 validation for float-like field types

codealchemy:
`float` and `decimal` fields currently render as text inputs, allowing non-numerical values in the form.

This updates their respective form helper to be the [appropriate field type (number)](https://apidock.com/rails/v5.2.3/ActionView/Helpers/FormHelper/number_field), as well as the [`step` html attribute](https://www.w3.org/TR/2012/WD-html5-20121025/common-input-element-attributes.html#attr-input-step) to the most permissive for input validation.

Closes https://github.com/sferik/rails_admin/issues/3289


mshibuya:
Excellent, thanks!

Issue #1852, opened 2013-12-04
undefined method `html_safe' for nil:NilClass


heygambo:
I've discovered an exception in rails_admin 0.6.0:

![screenshot 2013-12-04 21 24 09](https://f.cloud.github.com/assets/449438/1677003/69e717b0-5d22-11e3-90dd-56ba8dd8ba60.png)

Gemfile.lock:

```
GIT
  remote: git://github.com/ahoward/mongoid-grid_fs.git
  revision: 059012a2c2a9e0a5d6e67137752c3e918689c88a
  specs:
    mongoid-grid_fs (1.9.0)
      mime-types (~> 1.19)
      mongoid (>= 3.0, < 5.0)

GIT
  remote: git://github.com/evansagge/mongoid-rspec.git
  revision: cd7fa1ff285f73d0948876fb1250842718716a2f
  specs:
    mongoid-rspec (1.9.0)
      mongoid (>= 3.0.1)
      rake
      rspec (>= 2.14)

GIT
  remote: git://github.com/mongoid/mongoid.git
  revision: ec65b665430a13eab135fcd22b3b766bb8e8c8ea
  specs:
    mongoid (4.0.0)
      activemodel (~> 4.0.0)
      moped (~> 2.0.beta3)
      origin (~> 1.0)
      tzinfo (~> 0.3.37)

GIT
  remote: git://github.com/simi/mongoid-paranoia.git
  revision: e559a7ca265c83f351807d8c7bc35300b3e13b99
  specs:
    mongoid-paranoia (0.1)
      activemodel (~> 4.0.0)
      mongoid (~> 4.0)

GIT
  remote: git://github.com/smartinez87/exception_notification.git
  revision: 7c0f16375a6e56045bc21fb33d3e4ca09a69c6c4
  specs:
    exception_notification (4.0.1)
      actionmailer (>= 3.0.4)
      activesupport (>= 3.0.4)

GEM
  remote: https://rubygems.org/
  specs:
    actionmailer (4.0.1)
      actionpack (= 4.0.1)
      mail (~> 2.5.4)
    actionpack (4.0.1)
      activesupport (= 4.0.1)
      builder (~> 3.1.0)
      erubis (~> 2.7.0)
      rack (~> 1.5.2)
      rack-test (~> 0.6.2)
    activemodel (4.0.1)
      activesupport (= 4.0.1)
      builder (~> 3.1.0)
    activerecord (4.0.1)
      activemodel (= 4.0.1)
[...]

PR #3379, opened 2021-08-11
Update Boolean `pretty_value` to include default fallback

codealchemy:
I wasn't able to reproduce the 🐛 seen on #1852 on recent versions of `mongoid` / `MongoDB` (issue references `mongoid` v4, which was released roughly 7 years ago), so this may already be resolved in more recent versions of related dependencies. 

However, this change will at least prevent the interface from breaking in the event a boolean value doesn't match the explicit types checked in the `case` statement (closes https://github.com/sferik/rails_admin/issues/1852).

There's a related [Rubocop cop](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MissingElse) that would flag the missing `else` here as well (this cop is disabled by default).

### DIFF
--- a/lib/rails_admin/config/fields/types/boolean.rb
+++ b/lib/rails_admin/config/fields/types/boolean.rb
@@ -12,12 +12,12 @@ module RailsAdmin
 
           register_instance_option :pretty_value do
             case value
-            when nil
-              %(<span class='label label-default'>&#x2012;</span>)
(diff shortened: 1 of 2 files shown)


mshibuya:
Could you write a spec for this?

mshibuya:
Thanks!

PR #3382, opened 2021-08-16
Update file upload translations to avoid concatenation

### LINKED ISSUE #3260: Add own translation key for "Delete Image"
Currently the button to delete an image is solved by concatenating two translations:

https://github.com/sferik/rails_admin/blob/834c83da05f6194bf808f052e011d539ceaa3845/app/views/rails_admin/main/_form_file_upload.html.haml#L15

The problem is that in other languages, like German, the sequence of the words is grammatically wrong.

codealchemy:
This updates the file upload form translations to better align with i18n conventions (ie. avoid concatenating translations). 

Closes #3260

### DIFF
--- a/config/locales/rails_admin.en.yml
+++ b/config/locales/rails_admin.en.yml
@@ -127,6 +127,7 @@ en:
       save: "Save"
       save_and_add_another: "Save and add another"
       save_and_edit: "Save and edit"
+      delete_file: "Delete '%{field_label}' #%{number}"
       all_of_the_following_related_items_will_be_deleted: "? The following related items may be deleted or orphaned:"
       are_you_sure_you_want_to_delete_the_object: "Are you sure you want to delete this %{model_name}"
       confirmation: "Yes, I'm sure"
--- a/spec/integration/fields/multiple_carrierwave_spec.rb
+++ b/spec/integration/fields/multiple_carrierwave_spec.rb
@@ -30,7 +30,7 @@ RSpec.describe 'MultipleCarrierwave field', type: :request, active_record: true 
 
     it 'supports deleting a file', js: true do
       visit edit_path(model_name: 'field_test', id: field_test.id)
-      click_link 'Delete carrierwave assets #1'
+      click_link "Delete 'Carrierwave assets' #1"
       click_button 'Save'
       field_test.reload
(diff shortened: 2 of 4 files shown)


config/locales/rails_admin.en.yml
```diff
@@ -94,6 +94,7 @@ en:
         menu: "Delete"
         breadcrumb: "Delete"
         link: "Delete '%{object_label}'"
+        link_with_number: "Delete '%{object_label}' #%{object_number}"
```
mshibuya, line 97: Since this is not to be used in the Delete action, how do you think of creating this under `admin.form`?

```yaml
admin:
  form:
    delete_file: "Delete '%{field_label}' #%{number}"
```

spec/integration/fields/multiple_carrierwave_spec.rb
```diff
@@ -30,7 +30,7 @@
 
     it 'supports deleting a file', js: true do
       visit edit_path(model_name: 'field_test', id: field_test.id)
-      click_link 'Delete carrierwave assets #1'
+      click_link I18n.t('admin.actions.delete.link_with_number', object_label: 'Carrierwave assets', object_number: 1)
```
mshibuya, line 33: Can we restore this back to the original state? In this way we lose a way to check the content of I18n translation...

config/locales/rails_admin.en.yml
```diff
@@ -127,6 +127,7 @@ en:
       save: "Save"
       save_and_add_another: "Save and add another"
       save_and_edit: "Save and edit"
+      delete_file: "Delete '%{object_label}' #%{object_number}"
```
mshibuya, line 130: Those interpolation names are misleading. `object` in RailsAdmin stands for a model instance to be edited or deleted in member actions, but what's actually here is a field's label, not an object's label. Also `object_number` is not the number of an object, it's the number of an attached file.
Hence I propose `"Delete '%{field_label}' #%{number}"`.

mshibuya:
Thanks ❤️

config/locales/rails_admin.en.yml
codealchemy: Ah, I missed tweaking this after the first review - nice catch 🙇🏼

PR #3383, opened 2021-08-16
CI performance improvements

codealchemy:
This skips the intermediate `Cache gems` / `Install dependencies` steps by enabling [`bundler-cache` in the `setup-ruby` action](https://github.com/ruby/setup-ruby#caching-bundle-install-automatically).

Additionally, the added `MAKE` env var improves the performance of building native extensions on MRI (borrowing the approach mentioned [here](https://build.betterup.com/one-weird-trick-that-will-speed-up-your-bundle-install/))

This up to roughly 1m30s per MRI build in the current CI matrix (and low-to-no improvement for `jruby` builds)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -67,15 +67,11 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: ${{ matrix.ruby }}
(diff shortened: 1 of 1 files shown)


mshibuya:
Fantastic, thanks!

Issue #2893, opened 2017-06-10
Remove JQuery


prem-prakash:

mshibuya:
What's the motivation on this?

prem-prakash:
As Rails 5.1 arrived, and webpacker gem, we could manage assets using webpack and yarn.

But rails_admin is coupled with jQuery in a way that we could not listen to rails_admin events, like rails_admin.dom_ready event and others.

So it would be much more compatible with the new path that assets management is taking if rails_admin javascript could be vanilla instead of based on jQuery.

Taking a look at the code I could perceive that it would not so simple, because of pjax, that is a strong dependency and based on jQuery.

Maybe it is a good time to think about moving to Turbolinks or other library that can fulfil the same needs.

What do you think?

patleb:
I don't think the problem is jquery, pjax or any other js lib within this project, there is just no easy way to use webpacker and/or yarn with RailsAdmin since there is only support for sprockets.

rwdj:
I'd love to see a resolution to this. Currently using pure Webpacker and I'm unable to satisfactorily apply typescript to my view. None of the pjax events are firing, so the page has to be refreshed to fire `onload`.

PR #3387, opened 2021-09-15
Make colorpicker field use HTML5 native color picker

jdufresne:
This removes the jquery.colorpicker.js dependency, reducing the overall
size of compiled assets.

The native color input element is builtin and so will have better a11y
support. It allows the browser, device, or a11y technology to present a
UI to the user to select a color. The value submitted by the form is a 6
digit hex, same as before.

Beyond the advantage of using a native UI element, this helps move
towards removing jQuery as a dependency or at least reducing its usage.

For additional details on the color input element, see:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/color

Uses Rails color_field for rendering:
https://api.rubyonrails.org/classes/ActionView/Helpers/FormHelper.html#method-i-color_field

Refs #2893


mshibuya:
Amazing work, thanks!

PR #3388, opened 2021-09-15
Switch GitHub Actions to use ubuntu-latest

jdufresne:
Fixes the following error when running actions on a fork:

    Can't find any online and idle self-hosted runner in the current repository, account/organization that matches the required labels: 'ubuntu-16.04'
    Waiting for a self-hosted runner to pickup this job...


mshibuya:
👍

PR #3389, opened 2021-09-15
Remove unnecessary JavaScript dependency jquery.migrate.js

jdufresne:
The frontend JavaScript code is not using any jQuery API that requires
these compatibility shims. By removing the dependency we slightly reduce
the compiled asset size and work towards reducing overall jQuery usage.

Refs #2893


mshibuya:
> The frontend JavaScript code is not using any jQuery API that requires these compatibility shims.

How did you figure this out? It would be possible but I want to be careful since we have very limited test code coverage for JS logics...

jdufresne:
> How did you figure this out? It would be possible but I want to be careful since we have very limited test code coverage for JS logics...

Good question.

I installed this PR branch in my local project and clicked around as much as possible and looked for JavaScript console errors. There were none.

Additionally, I visually scanned the JavaScript code for known incompatibilities, I did not find any.

But I agree, this is not foolproof and there is still a possibility of bugs. If you find any or already know of some, please let me know and I'll spend some time fixing them.

If there are no known bugs, we basically have two options:

1. Keep `jquery.migrate.js` as it is uncertain whether or not bugs are introduced.

2. Drop `jquery.migrate.js`. If a related bug is reported, address it. I'm happy to help out on this front, if you @ me in a issue ticket.

Without eventually taking the plunge, I'm not sure of a good strategy for dropping the shim. If you decide the risk isn't worth it, no hard feelings. Please feel free to close.

mshibuya:
OK, I trust you 😃 
Thanks for the investigation!

PR #3390, opened 2021-09-17
Switch from jquery_ujs to rails-ujs

jdufresne:
rails-ujs has been available in Rails itself since 5.1.0:
https://github.com/rails/rails/tree/main/actionview/app/assets/javascripts

It obsoletes jquery_ujs. From the docs:
https://github.com/rails/jquery-rails

> If you are running Rails 5.1 and up, and if you have included //=
> require rails-ujs, then jquery_ujs is not needed anymore. You can just
> add:

    //= require jquery

This is a step towards reducing jQuery usage while still allowing for
UJS features.

Refs #2893


mshibuya:
Excellent, thanks!

PR #3392, opened 2021-09-18
Simplify jquery-ui inclusion

jdufresne:
Require rails-jquery-ui 6.0 (released Nov 29, 2016) which ships jQuery
UI 1.12.1 to allow simplifying by remove obsolete support.


mshibuya:
Thanks!

PR #3394, opened 2021-09-20
Add specs for JavaScript widgets

mshibuya:
Some examples are not stable enough...


mshibuya:
All steps are passing except for Coveralls ones, good to go!

PR #3396, opened 2021-09-22
Remove capitalization helper

codealchemy:
Added in https://github.com/sferik/rails_admin/commit/7f1033730dc948a12c86fd58e6a9e070f1b4dcfa, `#capitalize_first_letter` includes assumptions about letter order / casing for all strings, and may apply the wrong capitalization rules for some use cases / contexts (case-sensitive strings, different languages - related https://github.com/sferik/rails_admin/issues/2354, [StackOverflow Q/A](https://stackoverflow.com/questions/12169785/what-is-the-best-practice-to-capitalize-words-in-rails)).

Most of the use cases here are using translations, in which case the letter casing can / should be applied in the translations themselves. Other cases, such as rendering the value of `object_label_method`, will show the value as-is without transformation (as seen in the test updates). Cursory checks in the Dummy app show no changes to the UI.

My use case here involves the latter, where the object's label (case sensitive) in the breadcrumb navigation is capitalized and has led to some confusion.

The [CSS property `text-transform`](https://www.w3schools.com/cssref/pr_text_text-transform.asp) offers an alternative to achieve the same outcome, giving users of this library a way of applying letter-case logic in [theme overrides](https://github.com/sferik/rails_admin/wiki/Theming-and-customization).

### DIFF
--- a/app/helpers/rails_admin/application_helper.rb
+++ b/app/helpers/rails_admin/application_helper.rb
@@ -1,13 +1,5 @@
 module RailsAdmin
   module ApplicationHelper
-    def capitalize_first_letter(wording)
-      return nil unless wording.present? && wording.is_a?(String)
-
-      wording = wording.dup
-      wording[0] = wording[0].mb_chars.capitalize.to_s
-      wording
-    end
-
     def authorized?(action_name, abstract_model = nil, object = nil)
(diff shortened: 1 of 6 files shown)


mshibuya:
Makes sense, thanks!

PR #3398, opened 2021-09-22
Update `image?` check for file upload field types

codealchemy:
[Lookup the type ](https://apidock.com/rails/v6.0.0/Mime/Type/lookup_by_extension/class)via the [extension](https://apidock.com/rails/ActiveStorage/Filename/extension_without_delimiter), and match against 'image' types.

Rails provides a way to register custom mime types, allowing additional 'image' types to [be added to an Rails application](https://guides.rubyonrails.org/action_controller_overview.html#restful-downloads) without needing to maintain a list of specific filename extensions in `rails_admin`

ex. [`webp`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#webp) 

```ruby
# config/initializers/mime_types.rb
Mime::Type.register 'image/webp', :webp
```

Related update in Rails adding `webp` as a valid content type for `ActiveStorage` https://github.com/rails/rails/commit/2a0823ecbaeb3a6fbedc5cc31879d68fdf27d0cb 

Closes https://github.com/sferik/rails_admin/issues/3239

### DIFF
--- a/lib/rails_admin/config/fields/types/active_storage.rb
+++ b/lib/rails_admin/config/fields/types/active_storage.rb
@@ -17,7 +17,8 @@ module RailsAdmin
 
           register_instance_option :image? do
             if value
-              value.filename.to_s.split('.').last =~ /jpg|jpeg|png|gif|svg/i
+              mime_type = Mime::Type.lookup_by_extension(value.filename.extension_without_delimiter)
+              mime_type.to_s.match?(/^image/)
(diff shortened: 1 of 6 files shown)


mshibuya:
Could add some tests, like `webp` is considered as an image?

mshibuya:
Great, thanks!

PR #3400, opened 2021-10-11
Use `Kaminari.config#param_name` for `history_show` action

codealchemy:
Followup to https://github.com/sferik/rails_admin/pull/3227, this appears to be the last instance where `params[:page]` is referenced.


mshibuya:
Ah I overlooked, thanks!

PR #3401, opened 2021-10-11
Remove `tins` lock for Ruby 1.9

codealchemy:
Added in https://github.com/sferik/rails_admin/commit/e76440ebd15011f0a636b9dcc8d3cf618f567d56, support for Ruby < 2.1 was dropped [a while ago](https://github.com/sferik/rails_admin/blob/master/CHANGELOG.md#100rc---2016-07-18).


mshibuya:
Thanks 👍

PR #3402, opened 2021-10-12
Ensure `open` can be used as a field name

codealchemy:
Backfilling test coverage as requested in https://github.com/sferik/rails_admin/issues/1536#issuecomment-940697529, closes https://github.com/sferik/rails_admin/issues/1536


mshibuya:
Thanks! 🌟

PR #3406, opened 2021-10-23
Allow configuring multiple fields for a section at once

codealchemy:
This allows passing an array of field names to the `configure` block to simplify configuring multiple fields that share the same settings within a section.

Closes https://github.com/sferik/rails_admin/issues/2667


mshibuya:
Ah I didn't realize the `#fields`. That makes sense, thanks!

Issue #3404, opened 2021-10-15
Bulk export doesn't work with Pundit


goldaanton:
Hey, I'm using Pundit to authorize actions in Rails Admin. And recently I discovered, that I can't use bulk export for users, as Pundit tries to check unexpected policies.

Here are my steps to understand what happens:
1. Choose users and press `Export selected Users`.

![image](https://user-images.githubusercontent.com/48321400/137455910-55873510-fbdc-414b-b0ad-4abb12bead79.png)

2. And I get this error for `reset_user_password` action, which is really strange as I didn't use that action.

![image](https://user-images.githubusercontent.com/48321400/137454805-a061b852-02ac-4d1b-a64f-a2ec0f1efa80.png)

Thanks for any help!

Here is Full Trace:
`app/policies/admins/user_policy.rb:46:in 'reset_user_password?'
rails_admin (2.2.1) lib/rails_admin/extensions/pundit/authorization_adapter.rb:37:in 'authorized?'
rails_admin (2.2.1) lib/rails_admin/config/actions/base.rb:57:in 'block in <class:Base>'
rails_admin (2.2.1) lib/rails_admin/config/configurable.rb:75:in 'instance_eval'
rails_admin (2.2.1) lib/rails_admin/config/configurable.rb:75:in 'block in register_instance_option'
rails_admin (2.2.1) lib/rails_admin/config/configurable.rb:59:in 'block in register_instance_option'
rails_admin (2.2.1) lib/rails_admin/config/actions/base.rb:44:in 'block in <class:Base>'
rails_admin (2.2.1) lib/rails_admin/config/configurable.rb:75:in 'instance_eval'
rails_admin (2.2.1) lib/rails_admin/config/configurable.rb:75:in 'block in register_instance_option'
rails_admin (2.2.1) lib/rails_admin/config/configurable.rb:59:in 'block in register_instance_option'
rails_admin (2.2.1) lib/rails_admin/config/proxyable/proxy.rb:25:in 'method_missing'
rails_admin (2.2.1) lib/rails_admin/config/actions.rb:27:in 'select'
rails_admin (2.2.1) lib/rails_admin/config/actions.rb:27:in 'all'
rails_admin (2.2.1) app/controllers/rails_admin/main_controller.rb:28:in 'bulk_action'
actionpack (5.2.2.1) lib/action_controller/metal/basic_implicit_render.rb:6:in 'send_action'
[...]

PR #3407, opened 2021-10-26
Fix bug with MainController#bulk_action 

Art3606:
For checking if action exists, was used `:all` scope instead of `:bulkable`. Because of this, will-called all policy methods which possibly can raise some errors


app/controllers/rails_admin/main_controller.rb
```diff
@@ -25,7 +25,7 @@ def #{action.action_name}
     end
 
     def bulk_action
-      send(params[:bulk_action]) if params[:bulk_action].in?(RailsAdmin::Config::Actions.all(controller: self, abstract_model: @abstract_model).select(&:bulkable?).collect(&:route_fragment))
+      send(params[:bulk_action]) if params[:bulk_action].in?(RailsAdmin::Config::Actions.all(:bulkable, controller: self, abstract_model: @abstract_model).select(&:bulkable?).collect(&:route_fragment))
```
mshibuya, line 28: `.select(&:bulkable?)` is no longer necessary, right?

mshibuya:
Could you add a test for this fix?

mshibuya:
This shouldn't be considered as the fix for issue #3404 (just happens to go around it in your setup), but the fix itself makes sense.
After merging in I'll add a test, thanks.

PR #3409, opened 2021-11-01
Find-and-replace sferik owner with railsadminteam

ryanwilsonperkin:
Noticed that some of the links were broken because they pointed to the
repo with sferik as owner, but the repo has been transfered to
railsadminteam.

GitHub generally does a pretty good job of forwarding these along so the
links for issues/pull requests still worked, but references to
individual commits or files were broken.

For example, the [d72090e](https://github.com/sferik/rails_admin/commit/d72090ec6a07c3b9b7b48ab50f3d405f91ff4375) link in the README does not resolve


mshibuya:
Thanks ❤️

PR #3415, opened 2021-11-14
Add Prettier to auto-format JavaScript and more

jdufresne:
https://github.com/prettier/prettier

In many ways, Prettier is like RuboCop but for JavaScript.

Prettier is an opinionated code formatter. It enforces a consistent
style to make new contributions easier. Contributors do not need to
concern themselves with the project's styling preferences, just let the
tool handle it. Reviewers don't need to enforce a personal style guide,
GitHub actions will automatically check all pull requests.

Vendored code is excluded from the formatter through the user of the
.prettierignore file.


mshibuya:
Great, thanks!

PR #3416, opened 2021-11-14
Remove nearly all jQuery from ra.sidescroll.js

jdufresne:
All DOM methods now use native browser features.

The only remaining jQuery is used to listen to the custom jQuery event
"rails_admin.dom_ready".

Refs #2893


mshibuya:
Thanks!

PR #3418, opened 2021-11-14
Cleanup several RuboCop TODO items and regenerate file

jdufresne:


jdufresne:
Rebased to resolve conflicts and regenerated the todo file. CI is now green.

app/controllers/rails_admin/main_controller.rb
```diff
@@ -54,21 +54,19 @@ def get_sort_hash(model_config)
...
+                 "#{abstract_model.table_name}.#{model_config.list.sort_by}"
+               elsif (field.sortable.is_a?(String) || field.sortable.is_a?(Symbol)) && field.sortable.to_s.include?('.') # just provide sortable, don't do anything smart
+                 field.sortable
+               elsif field.sortable.is_a?(Hash) # just join sortable hash, don't do anything smart
+                 "#{field.sortable.keys.first}.#{field.sortable.values.first}"
+               elsif field.association? # use column on target table
+                 "#{field.associated_model_config.abstract_model.table_name}.#{field.sortable}"
+               else # use described column in the field conf.
+                 "#{abstract_model.table_name}.#{field.sortable}"
+               end
```
mshibuya, line 69: How about adding this to .rubocop.yml:

```yaml
Layout/MultilineAssignmentLayout:
  Enabled: true
```

so this code becomes like this?

```ruby
      column =
        if field.nil? || field.sortable == true # use params[:sort] on the base table
          "#{abstract_model.table_name}.#{params[:sort]}"
        elsif field.sortable == false # use default sort, asked field is not sortable
          ...
```

jdufresne: I'm not against adding that, but it is not my first preference.

When faced with a formatting choice, I tend to prefer to rely on defaults (usually, but not always) and defer to documneted community standards.

In this particular case, the new style would go against the Ruby style guide: https://rubystyle.guide/#indent-conditional-assignment

The guide suggests to use the format in this PR which is also the RuboCop default.

All that said, it is just a preference. I'm happy to add it regardless as it is not a blocker for me. Let me know how you feel, I'll defer the final decision to you.

mshibuya: Thanks for your opinion, then please add it.
My point is that to comply with the existing style if it's reasonable is more important than to rely on defaults. I don't want unnecessary(as you said, it's just a preference) diffs to show up when I do git blame to investigate issues.

> the new style would go against the Ruby style guide

This does not `go against` it, since this style is also stated as

```
# good (and a bit more width efficient)
kind =
  case year
  when 1850..1889 then 'Blues'
  ...
```

jdufresne: :+1: Done in the latest revision.

mshibuya:
Squash-merged this, thanks!

PR #3419, opened 2021-11-16
✨ Add target_window to actions

RocKhalil:
Added the ability to change the target window of the action.
For example, I want to open an `Import` page in a different window, I'll need to do:

`target_window :_blank`


RocKhalil:
@jdufresne @mshibuya not sure why I'm marked as a first time contributor, but anyway, can you guys check this please ?

lib/rails_admin/config/actions/base.rb
```diff
@@ -78,6 +78,11 @@ class Base
           true
         end
 
+        # Target window [_self, _blank]
+        register_instance_option :target_window do
+          :_self
```
jdufresne, line 83: Could we default to no attribute instead? Maybe that would be handled by `nil`? In the normal cause I wouldn't expect `target="_self"`, just nothing.

jdufresne (commented):
would it be possible to add a test showing the rendered result?

lib/rails_admin/config/actions/base.rb
RocKhalil: Removed the default value, however, I'm not sure what's the test that can cover this change;
Any idea ?

jdufresne: >  Any idea ?

Would id be possible to write a test that renders HTML with this option? This HTML either contains the attribute `target="_blank"` or it doesn't.

RocKhalil:
@jdufresne I've added a test, however, I'm not able to install `mysql` gem on my mac for some reason, so I wasn't able to run it to validate it.
Can you please double check ?

lib/rails_admin/config/actions/base.rb
```diff
@@ -78,6 +78,11 @@ class Base
           true
         end
 
+        # Target window [_self, _blank]
+        register_instance_option :target_window do
```
mshibuya, line 82: To me, `link_target` feels like to be a better name. What do you think?

lib/rails_admin/config/actions/base.rb
```diff
@@ -78,6 +78,11 @@ class Base
           true
```
mshibuya, line 78:
```suggestion
          target_window.blank?
```
Or maybe `target_window != :_blank`

app/helpers/rails_admin/application_helper.rb
```diff
@@ -163,7 +163,14 @@ def menu_for(parent, abstract_model = nil, object = nil, only_icon = false)
...
               end
-            content_tag(:a, label, {href: href}.merge(action.pjax? ? {class: ['pjax']} : {}))
+
+            extra_classes = []
+            extra_classes << 'pjax' if action.pjax? && action.target_window != :_blank
+
+            extra_tags = {}
+            extra_tags[:target] = action.target_window if action.target_window.present?
+
+            content_tag(:a, label, {href: href, class: extra_classes}.merge(extra_tags))
```
mshibuya, line 173:
```suggestion
            content_tag(:a, label, {href: href, target: action.target_window}.merge(action.pjax? ? {class: ['pjax']} : {}))
```

RocKhalil: One of the issues of PJAX, is that, when enabled, it was removing the effect of the `target=_blank` for some reason

lib/rails_admin/config/actions/base.rb
RocKhalil: Yes, makes sense! why not

mshibuya:
Merged in, thanks for the contribution 🙌

eliten00b:
Can we have this back ported and release on version 2.x?

PR #3422, opened 2021-11-20
Remove unnecessary IIFE from ra.sidescroll.js

jdufresne:
The file no longer uses jQuery, so it doesn't need to be passed to the
IIFE (Immediately Invoked Function Expression).

The IIFE can be replaced by a simple block to continue to ensure the
global namespace is not polluted.

Refs #2893


mshibuya:
Thanks!

PR #3423, opened 2021-11-20
Remove unnecessary parentheses around keyword `raise`

jdufresne:
`raise` is a keyword, not a function. In general, it should not have
parentheses unless they are required for expression evaluation or
precedence. These have been removed to better match the examples in the
Ruby style guide.

https://rubystyle.guide/#exceptions


mshibuya:
👍

PR #3424, opened 2021-11-20
Only run coveralls on the base repository

jdufresne:
Allow contributors to run GitHub actions on their fork successfully.
Before this changed, local actions always fail on the coverall step as
the secret is unavailable. Now it only runs on the base
repository (railsadminteam/rails_admin).


mshibuya:
Cool, thanks!

PR #3425, opened 2021-11-22
Switch templates from HAML to ERB

### LINKED ISSUE #3173: Feature Request: use ERB instead of HAML
TL;DR: This Issue is about using the preinstalled ERB instead of HAML for templating.

@mroderick and I are Rails Admin users. We'd volunteer to try to make this happen, provided this Issue is accepted.

### The landscape

- Rails Admin was created in a cultural moment when HAML was everywhere
- Time passed, HAML lost its dominance
- People have asked "Can we lose HAML?" before, and then was not the time
- Rails Admin 2.0 could have fewer opinions. Perhaps now is the time.

### Reasons to allow people not to use HAML

- Users and maintainers get to have **fewer gem dependencies**
- New users won't have to learn HAML, everyone has to learn ERB

## Solution proposal

- before 2.0, change all templates that come with Rails Admin to ERB
- have a tool-assisted workflow, to make this feasible
    - https://github.com/danchoi/herbalizer
- try to keep people's apps working, with their existing template overrides

## Elsewhere

* https://thoughtbot.com/blog/fashion-runway-erb-vs-haml
* https://opensoul.org/2011/11/30/haml-the-unforgivable-sin/
* https://speakerdeck.com/klaustopher/erb

jdufresne:
Make RailsAdmin have fewer dependencies and so slightly reduce total
installation size.

Align the project with default upstream Rails practices. Starting a new
Rails project includes ERB by default but not HAML. By switching, we
don't pull in additional templating practices beyond what any default
project is already using.

The change may also help encourage contributors. While nearly all Rails
developers have direct experience working with ERB, the same is not
always true of HAML. This removes a small barrier.

Fixes #3173


mshibuya:
Let's go with this, thanks!

mroderick:
@jdufresne 🌟

PR #3430, opened 2021-11-28
Remove support for EOL Ruby 2.5

jdufresne:
Ruby 2.5 went EOL 2021-03-31. It is no longer supported upstream and so
is not receiving bug fixes, including for security issues. Reduce the
testing resources on RailsAdmin.

https://www.ruby-lang.org/en/downloads/branches/

Dropping EOL Ruby versions allows the project to begin taking advantage
of newer syntax and library features.

TargetRubyVersion was removed from .rubocop.yml as this value is
determined automatically from the gemspec.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
         orm: [active_record]
         adapter: [sqlite3]
         include:
-          - ruby: 2.5
+          - ruby: 2.6
             gemfile: gemfiles/rails_6.0.gemfile
             orm: active_record
             adapter: sqlite3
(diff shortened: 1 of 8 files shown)


mshibuya:
Though being not documented yet (when this discussion is settled I will do it soon), existing way of deciding supported version is:

- Support 2 major version of Rails (e.g. RailsAdmin 3.x supports Rails 6.x and 7.x)
- Support Ruby version which are allowed in the lowest Rails version (e.g. Rails 6.0 supports Ruby >= 2.5, since use >= 2.5 for RailsAdmin 3.x)

How do you think of keeping this approach or do you have a different opinion?

> Dropping EOL Ruby versions allows the project to begin taking advantage of newer syntax and library features.

I really, really understand this point, but we need to think about existing users who can't upgrade their app easily.

mshibuya:
Merging in, since I realized that to introduce Turbo Drive we need to drop support for Ruby 2.5.
Thank you for the contribution!

PR #3434, opened 2021-11-29
Enable rubocop-performance

jdufresne:
The gem was included in the Gemfile, just not enabled in the RuboCop
configuration.


mshibuya:
Great finding!

PR #3435, opened 2021-11-29
Remove JavaScript library jquery-pjax

jdufresne:
The library is no longer maintained and has not received a new release
since May 20, 2017. See https://github.com/defunkt/jquery-pjax

> Status of this project
>
> jquery-pjax is largely unmaintained at this point.

This also removes the Ruby gem rack-pjax which is also no longer
maintained and has not received a new release since Mar 24, 2019.

This is work towards the goal of dropping jQuery in favor of more
vanilla JavaScript solutions. As a dependency using jQuery, pjax was a
barrier toward this goal.

jquery-pjax was originally introduced in
645d678732eb550db2c5b250aa16f5ac5e9f8d62 (Nov 16, 2011). Presumably as a
page loading optimization, however it is not strictly necessary to
achieve the set of RailsAdmin features. After many years of improvements
in browser caching, loading, and JavaScript evaluation it is unclear if
this optimization is still necessary for a good UX.

Refs #2893


mshibuya:
>  it is unclear if this optimization is still necessary for a good UX

This is understandable, but I'm not totally sure about it. It would be ideal if we can perform quantitative comparison...

Another solution will be migrating to [Turbo Drive](https://turbo.hotwired.dev/handbook/drive), which I understand is modernized version of Pjax.

mshibuya:
Squash-merged with 44451f27e10d65e6b16737e17f81f9578ee7f752, thanks!

I'll try introducing Turbo Drive later.

mshibuya:
@jdufresne Sorry I failed to put your credit to the commit when I handled squash merge locally.
It was unintentional and I apologize 🙇

jdufresne:
No worries at all. Thanks for merging and for Rails admin in general!

PR #3437, opened 2021-11-29
Fix #3420

### LINKED ISSUE #3420: V3.0.0.alpha : running in production is breaking
For some reason, customizing actions ( or even including them ) in `initializers/rails_admin.rb` is breaking;

```rb
def add_action_custom_key(action, &block)
  action.instance_eval(&block) if block
  @@actions ||= []
  if action.custom_key.in?(@@actions.collect(&:custom_key))
    raise "Action #{action.custom_key} already exists. Please change its custom key."
  else
    @@actions << action
  end
end
```

Thrown error:
```
add_action_custom_key: Action dashboard already exists. Please change its custom key. (RuntimeError)
```

RocKhalil:
@mshibuya You forgot to open a PR for this;
It's ready to merge, we've been testing it for a week now on production

### DIFF
--- a/lib/rails_admin/engine.rb
+++ b/lib/rails_admin/engine.rb
@@ -27,7 +27,9 @@ module RailsAdmin
     initializer 'RailsAdmin reload config in development' do |app|
       config.initializer_path = app.root.join('config/initializers/rails_admin.rb')
 
-      unless Rails.application.config.cache_classes
+      if Rails.application.config.cache_classes
+        RailsAdmin::Config.initialize!
+      else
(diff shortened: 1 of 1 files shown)


mshibuya:
Thanks for opening a PR, but that fix is kind of a workaround. I need to make sure that the solution is proper one, since Rails initialization process is complex...
https://guides.rubyonrails.org/initialization.html

RocKhalil:
Yes, totally get that.
However, we've been using it for the past 2 weeks in a production environment and there are no issue with it

PR #3439, opened 2021-11-29
🐛 Fix boolean field

RocKhalil:
The boolean field was rendering in a wrong way

### DIFF
--- a/app/views/rails_admin/main/_form_boolean.html.erb
+++ b/app/views/rails_admin/main/_form_boolean.html.erb
@@ -1,7 +1,7 @@
 <% if field.nullable? %>
   <div class="btn-group" data-toggle="buttons">
     <% {'1': true, '0': false, '': nil}.each do |text, value| %>
-      <label class="<%= [field.css_classes[value], ("active" if field.form_value == value)] %> btn btn-default">
+      <label class="<%= [field.css_classes[value], ("active" if field.form_value == value)].compact.join(' ') %> btn btn-default">
         <%= form.radio_button field.method_name, text, field.html_attributes.reverse_merge({ checked: field.form_value == value, required: field.required}) %>
         <%= field.labels[value].html_safe %>
       </label>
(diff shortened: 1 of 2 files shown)


app/views/rails_admin/main/_form_boolean.html.erb
```diff
@@ -1,7 +1,7 @@
 <% if field.nullable? %>
   <div class="btn-group" data-toggle="buttons">
     <% {'1': true, '0': false, '': nil}.each do |text, value| %>
-      <label class="<%= [field.css_classes[value], ("active" if field.form_value == value)] %> btn btn-default">
+      <label class="<%= [field.css_classes[value], ("active" if field.form_value == value)].compact.join(' ') %> btn btn-default">
```
jdufresne, line 4: Could a test be added (or extended) to demonstrate the fix and would have caught the regression?

RocKhalil: I think the regression was caused when we decided to add the `nil` field and change the layout ( mainly with haml integration ).
I don't think there's a necessity to amend the tests as this was a UI bug ( not sure if there are UI tests )

jdufresne:
> I think the regression was caused when we decided to add the nil field and change the layout ( mainly with haml integration ).
>
> I don't think there's a necessity to amend the tests as this was a UI bug ( not sure if there are UI tests )

A rendering bug is still a bug that can be caught by CI. I don't understand why we should have a policy against such tests if they help refactoring code and catching regressions. Basically, the `<label>` element is expected to render with the class element formatted a certain way, if it is not, that is a bug that can be tested.

RocKhalil:
yup @jdufresne , I removed the old comment because it made sense to fix the tests;
I updated them and they are now testing the `.success.active`, etc.. cases to make sure they are working correctly

jdufresne (approved):
Thanks!

Issue #3440, opened 2021-11-29
Use Or instead of And with multiple search conditions


QPierre1:
How is it possible to search in a rails admin list with "OR" multiple conditions ?

It possible to add conditions but only with "AND".

codealchemy:
The open [`Search` box](https://github.com/railsadminteam/rails_admin/wiki/List#fields-searching) will use `OR` to join across configured fields. You can also [customize the search for a configured model](https://github.com/railsadminteam/rails_admin/wiki/Custom-Search).

https://github.com/railsadminteam/rails_admin/blob/c7f907c0de1afa0f948807b2851dd8487976ed41/lib/rails_admin/adapters/active_record.rb#L130-L140

Does this help address what you're looking for?

QPierre1:
@codealchemy Thank you for your answer.

I'm concerning about how to join filters and not the search box.
OR is used to join across configured fields but AND is used if we add filters with the same field multiple times.

Here is an example of input that will use AND :
[https://media.discordapp.net/attachments/436982626763669524/915892578212282428/unknown.png](url)

I think OR would be more appropriate in this case. It's impossible to a field to be exactly two different things. I open https://github.com/railsadminteam/rails_admin/pull/3442 about that.

rnevius:
Has this functionality been added? Double checking before I write my own methods.

mshibuya:
It's not added, that's why this issue is still open. Attempts were made like https://github.com/railsadminteam/rails_admin/pull/3445 and https://github.com/railsadminteam/rails_admin/pull/3696, but we're yet to solve this problem cleanly enough.

PR #3442, opened 2021-12-01
evol:use and for distinct keys and or for the same keys

QPierre1:
When a user adds filters for the same field, use "or" instead of "and" with active_record and mongoid.
https://github.com/railsadminteam/rails_admin/issues/3440


mshibuya:
I'm sorry but I don't support this change. Same-looking filter entries magically changes its behavior from `AND` to `OR` condition is confusing enough for end-users.
Of course I understand the needs for OR filtering, but it should be achieved with a UI that's not confusing.

PR #3445, opened 2021-12-07
evol:add or/and separator field for text field

Quent1Pr:
Add and/or seperator with UI


mshibuya:
Very sorry for saying this after the long wait, but I came to conclude that this implementation is complicated, compared with what it aims to achieve. Especially mixture of "and" and "or" has edge-cases, which are hard to handle in a intuitive way.
I've prepared a PR to apply your change into the current codebase. If we were to revisit the decision it'll be the starting point.
#3696

Issue #2790, opened 2016-12-09
config.parent_controller setting does not work


grillermo:
On my project the config/initializers/rails_admin.rb file added 

    config.parent_controller = '::MyBaseController'

And rails admin ignores it, if i change directly the lib/rails_admin/config.rb file and hardcode this

    @parent_controller = '::MyBaseController

It works, my conclusion is that this mechanism 

    class ApplicationController < Config.parent_controller.constantize

Doesnt work.
I tested it using a before_action filter and i could confirm that doing

    self.class.superclass

Does not return the right class when using the config approach.

tamaloa:
Had a similar problem when trying to set `config.parent_controller = '::ApplicationController'` to keep the old behaviour.
Could fix it by moving the config.parent_controller setting above my `config.audit_with` call. Try moving it to the very beginning of the config block. I guess some config calls instantiate the controller so that the setting later takes no effect anymore.

grillermo:
Will try that, thank you

PR #3458, opened 2022-01-24
Fix inheritance of parent_controller when controllers were eagerly loaded

jalyna:
Follow up on https://github.com/railsadminteam/rails_admin/issues/2790

https://github.com/railsadminteam/rails_admin/commit/5bd980564a565906a6fda87d2ef31590d3e3b0a5 fixes the inheritance of the `RailsAdmin::ApplicationController`, nevertheless `RailsAdmin::MainController.superclass.superclass` is `ActionController::Base` instead of the configured `::ApplicationController`, when accessing the main controller first (happens with eager loading).

Before the fix:

```
  1) RailsAdmin::Config.parent_controller= if RailsAdmin::ApplicationController is already loaded can be changed
     Failure/Error: expect(RailsAdmin::MainController.superclass.superclass).to eq ApplicationController

       expected: ApplicationController
            got: ActionController::Base

       (compared using ==)

       Diff:
       @@ -1 +1 @@
       -ApplicationController
       +ActionController::Base
```

### DIFF
--- a/lib/rails_admin/config.rb
+++ b/lib/rails_admin/config.rb
@@ -269,9 +269,15 @@ module RailsAdmin
 
       def parent_controller=(name)
         @parent_controller = name
+
         if defined?(RailsAdmin::ApplicationController)
           RailsAdmin.send(:remove_const, :ApplicationController)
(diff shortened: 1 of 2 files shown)


mshibuya:
Awesome, thanks!

PR #3461, opened 2022-01-30
Introduce Turbo Drive

mshibuya:
Refs. https://github.com/railsadminteam/rails_admin/pull/3435#issuecomment-983566296

To test the significance of the performance improvement made by Turbo Drive, I performed a benchmark simulating site navigation with and without Turbo Drive.

### Preconditions
- The benchmark was performed using Chrome DevTools Performance panel, with CPU throttled to `6x slowdown` to make significance.
- Ran [dummy_app](https://github.com/railsadminteam/rails_admin/tree/master/spec/dummy_app) locally, with Webpack assets precompiled and CSSes extracted
- To test both page load and navigation in a time-precise way, these snippets are added to `src/rails_admin/ui.js`:
  - For non Turbo Drive version:
      ```diff
         $(document).ready(function () {
           I18n.init($("html").attr("lang"), $("#admin-js").data("i18nOptions"));

           const event = new CustomEvent("rails_admin.dom_ready");
           document.dispatchEvent(event);
         });
      +  window.addEventListener('load', function() {
      +    window.setTimeout(function () {
      +      if (window.performance.getEntriesByType('navigation').map((nav) => nav.type).includes('reload') && window.location.pathname == '/admin/') {
      +        window.location.href = '/admin/player';
      +      } else if (window.location.pathname == '/admin/player') {
      +        window.location.href = '/admin/player/752/edit';
      +      }
      +    }, 200);
      +  });
      ```
  - For Turbo Drive version:
      ```diff
         document.addEventListener("turbo:load", function () {
           I18n.init($("html").attr("lang"), $("#admin-js").data("i18nOptions"));
       
           const event = new CustomEvent("rails_admin.dom_ready");
           document.dispatchEvent(event);
         });
      +  window.addEventListener('load', function() {
      +    window.setTimeout(function () {
      +      if (window.location.pathname == '/admin/') {
      +        window.Turbo.visit('/admin/player');
      +      }
      +    }, 200);
      +  });
      +  window.addEventListener('turbo:render', function() {
      +    window.setTimeout(function () {
      +      if (window.location.pathname == '/admin/player') {
      +        window.Turbo.visit('/admin/player/752/edit');
      +      }
      +    }, 200);
      +  });
      ```

### Result

(I tried my best to roughly align the charts to represent the same timeframe per width, though it may be somewhat inaccurate...)

#### Without Turbo Drive
image

#### With Turbo Drive
image

I can roughly say that while the benchmark without Turbo Drive took roughly 1.1s to finish rendering the `edit` page, Turbo Drive one was reduced to 0.9s. In a real usage there will be more page transitions than the initial page load, so this difference will be more significant.

Based on this, my assumption is that introducing Turbo Drive will improve the UX.

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -35,6 +35,7 @@ appraise 'rails-6.1' do
   gem 'rails', '~> 6.1.0'
   gem 'sassc-rails', '~> 2.1'
   gem 'devise', '~> 4.7'
+  gem 'turbo-rails', platform: :jruby, github: 'mshibuya/turbo-rails', branch: 'fix/breaking-in-non-zeitwerk-env'
 
   group :test do
     gem 'cancancan', '~> 3.2'
--- a/app/assets/javascripts/rails_admin.js
+++ b/app/assets/javascripts/rails_admin.js
@@ -1,4 +1,5 @@
 //=  require 'rails-ujs'
+//=  require 'turbo'
 //=  require 'rails_admin/jquery3'
 //=  require 'jquery_nested_form'
 //=  require 'rails_admin/jquery-ui/effect'
--- a/gemfiles/rails_6.1.gemfile
+++ b/gemfiles/rails_6.1.gemfile
@@ -8,6 +8,7 @@ gem "rails", "~> 6.1.0"
 gem "webpacker", require: false
 gem "webrick", "~> 1.7"
 gem "sassc-rails", "~> 2.1"
+gem "turbo-rails", platform: :jruby, github: "mshibuya/turbo-rails", branch: "fix/breaking-in-non-zeitwerk-env"
 
 group :active_record do
   gem "paper_trail", ">= 12.0"
--- a/lib/rails_admin/config/fields/types/action_text.rb
+++ b/lib/rails_admin/config/fields/types/action_text.rb
@@ -9,7 +9,7 @@ module RailsAdmin
           RailsAdmin::Config::Fields::Types.register(self)
 
           register_instance_option :version do
-            '1.1.1'
+            '1.3.1'
           end
 
           register_instance_option :css_location do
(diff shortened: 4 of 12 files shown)


mshibuya:
Specs are still a bit unstable, but the failures looks like random ones.

mshibuya:
Merging this for now. Feel free to leave comments if there's anything to discuss!

codealchemy:
📝  This looks to only work if `zeitwerk` is enabled, as [`Rails.autoloaders.once`](https://github.com/hotwired/turbo-rails/blob/6958e0e2742d4744da80128d178e7255989386a4/lib/turbo/engine.rb#L19-L21) only returns a `Zeiterk::Loader` ([for Rails 6.0 / 6.1, at least](https://apidock.com/rails/v6.0.0/Rails/Autoloaders/once/class))

Updating a Rails 6.0 project that uses the `classic` autoloader fails to boot with the error

```
/bundle/ruby/2.7.0/gems/turbo-rails-1.0.1/lib/turbo/engine.rb:20:in `block in <class:Engine>': undefined method `do_not_eager_load' for nil:NilClass (NoMethodError)
```

mshibuya:
Yeah it is, I should have mentioned it here 🙇‍♂️

lrusso96:
@mshibuya even with the latest release candidate (rc2) I receive the following `Error: Form responses must redirect to another location` when tying to export to CSV/Json/XML.
[Here](https://stackoverflow.com/questions/70400958/error-form-responses-must-redirect-to-another-location) they suggest to disable Turbo for this kind of scenario.

Am I missing anything or should I open a new issue for this?

mshibuya:
Thanks for letting me know 👍  , it was fixed in https://github.com/railsadminteam/rails_admin/commit/c749d939e29434937e8558bcb1f3e219fe98c69d.

PR #3470, opened 2022-02-14
[ActiveRecord] Fix DB connection issues during `assets:precompile`

### LINKED ISSUE #3469: 3.0.0.rc try to access to the database when asset precompiling if it has config.model configuration
**Describe the bug**

If the Rails Admin's configuration file `config/initializers/rails_admin.rb` have `config.model` entry in it,
this rails application tries to access to the database when executing `bin/rails assets:precomplie` command. 
(If no any `config.model` entries, it won't access to the database at all.)

So if there is no DB available, it fails to precompile. This is the output:

```
$ bin/rails assets:precompile
rails aborted!
ActiveRecord::ConnectionNotEstablished: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory
        Is the server running locally and accepting connections on that socket?
/home/yuta/src/private/rails/rails-admin-eval/config/initializers/rails_admin.rb:43:in `block in <main>'
/home/yuta/src/private/rails/rails-admin-eval/config/environment.rb:5:in `<main>'

Caused by:
PG::ConnectionBad: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory
        Is the server running locally and accepting connections on that socket?
/home/yuta/src/private/rails/rails-admin-eval/config/initializers/rails_admin.rb:43:in `block in <main>'
/home/yuta/src/private/rails/rails-admin-eval/config/environment.rb:5:in `<main>'
Tasks: TOP => environment
(See full trace by running task with --trace)
```

**Reproduction steps**

1. Create a new Rails application with PostgreSQL database and create a model.
2. Install and setup Rails Admin following the installation process written on the README.
3. `bin/rails s` to confirm that Rails Admin works nicely.
4. Stop PostgreSQL server.
5. Confirm `bin/rails assets:precompile` works.
6. Add some `config.model` block to `config/initializers/rails_admin.rb` like following:
    ```
    RailsAdmin.config do |config|
      
      ...
      
      config.model "ModelYouHaveCreated" do
        visible true
      end
    end
    ```
7. Confirm `bin/rails assets:precompile` fails.

**Expected behavior**

Asset precompile process completes without accessing to the DB, because Rails Admin 2.2.1 didn't access to the DB even in the same conditions.

(This behavior on 3.0.0.rc would cause a problem for my application's deployment process, which should be done on CI without DB.)

**Additional context**

pg 1.3.1
rails 6.1.4.6
rails_admin 3.0.0.beta2, 3.0.0.rc and current master (3d7f3b3), with `config.asset_source = :webpacker` in `config/initializers/rails_admin.rb`
webpacker 5.4.3

and with PostgreSQL 13 for DB
(* Sorry but I haven't checked if this behavior occurs with other databases like MySQL or not.)

codealchemy:
Closes #3469

When initializing abstract [(`ActiveRecord`) models](https://github.com/railsadminteam/rails_admin/blob/3d7f3b33a1ca6fabbd8606bb178babae930cce25/lib/rails_admin/abstract_model.rb#L49), `#table_exists?` is used and [calls through the `connection`](https://github.com/rails/rails/blob/086935ea1db543037609cd33f3b4d29e9fdd5c9a/activerecord/lib/active_record/model_schema.rb#L393-L396), which (in the case of a `precompile` step) can raise errors other than the `ActiveRecord::NoDatabaseError` that's [currently handled](https://github.com/railsadminteam/rails_admin/blob/3d7f3b33a1ca6fabbd8606bb178babae930cce25/lib/rails_admin/abstract_model.rb#L22-L25).

I believe this should cover all [PostgreSQL](https://github.com/rails/rails/blob/ffb3a3e01a17de4be457437443dcc7163d0330fe/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L79-L89) and [MySQL](https://github.com/rails/rails/blob/ffb3a3e01a17de4be457437443dcc7163d0330fe/activerecord/lib/active_record/connection_adapters/mysql2_adapter.rb#L45-L55) connection errors handled by Rails.

### DIFF
--- a/lib/rails_admin/abstract_model.rb
+++ b/lib/rails_admin/abstract_model.rb
@@ -19,7 +19,7 @@ module RailsAdmin
       def new(m)
         m = m.constantize unless m.is_a?(Class)
         (am = old_new(m)).model && am.adapter ? am : nil
-      rescue *([LoadError, NameError] + (defined?(ActiveRecord) ? ['ActiveRecord::NoDatabaseError'.constantize] : []))
+      rescue *([LoadError, NameError] + (defined?(ActiveRecord) ? ['ActiveRecord::NoDatabaseError'.constantize, 'ActiveRecord::ConnectionNotEstablished'.constantize] : []))
         puts "[RailsAdmin] Could not load model #{m}, assuming model is non existing. (#{$ERROR_INFO})" unless Rails.env.test?
         nil
       end
--- a/spec/rails_admin/abstract_model_spec.rb
+++ b/spec/rails_admin/abstract_model_spec.rb
@@ -21,6 +21,16 @@ RSpec.describe RailsAdmin::AbstractModel do
         expect(RailsAdmin::AbstractModel.new('WithoutTable')).to eq nil
       end
     end
+
+    context 'on ActiveRecord::ConnectionNotEstablished', active_record: true do
+      before do
+        expect(WithoutTable).to receive(:table_exists?).and_raise(ActiveRecord::ConnectionNotEstablished)
+      end
+
+      it 'does not raise error and returns nil' do
+        expect(RailsAdmin::AbstractModel.new('WithoutTable')).to eq nil
+      end
+    end
   end
 
   describe '#to_s' do


mshibuya:
Looks great, thanks 🙌

ha4gu:
Now `bin/rails assets:precompile` works, thank you so much!

PR #3475, opened 2022-02-22
Fix: edit_user_link rendering

thedanielhanke:
if `config.show_gravatar` is e.g. `false`, "false" becomes part of the render output as `false` is not nil and therefore not removed on `#compact`

if there is no edit-user action to be found, the rendered output was 2 nested spans, which messed up layout.

### DIFF
--- a/app/helpers/rails_admin/application_helper.rb
+++ b/app/helpers/rails_admin/application_helper.rb
@@ -20,17 +20,24 @@ module RailsAdmin
     end
 
     def edit_user_link
-      return nil unless _current_user.respond_to?(:email)
-      return nil unless abstract_model = RailsAdmin.config(_current_user.class).abstract_model
+      return nil unless _current_user.try(:email).present?
+      return nil unless (abstract_model = RailsAdmin.config(_current_user.class).abstract_model)
 
-      content = [
-        RailsAdmin::Config.show_gravatar && _current_user.email.present? && image_tag("#{request.ssl? ? 'https://secure' : 'http://www'}.gravatar.com/avatar/#{Digest::MD5.hexdigest _current_user.email}?s=30", alt: ''),
-        content_tag(:span, _current_user.email),
-      ].compact.join.html_safe
-      if (edit_action = RailsAdmin::Config::Actions.find(:edit, controller: controller, abstract_model: abstract_model, object: _current_user)).try(:authorized?)
-        link_to content, rails_admin.url_for(action: edit_action.action_name, model_name: abstract_model.to_param, id: _current_user.id, controller: 'rails_admin/main'), class: 'nav-link'
+      edit_action = action(:edit, abstract_model, _current_user)
+      authorized = edit_action.try(:authorized?)
+      content = edit_user_link_label
+
+      if authorized
+        edit_url = rails_admin.url_for(
+          action_name: edit_action.action_name,
+          model_name: abstract_model.to_param,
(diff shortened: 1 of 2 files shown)


app/helpers/rails_admin/application_helper.rb
```diff
@@ -26,11 +26,11 @@ def edit_user_link
       content = [
         RailsAdmin::Config.show_gravatar && _current_user.email.present? && image_tag("#{request.ssl? ? 'https://secure' : 'http://www'}.gravatar.com/avatar/#{Digest::MD5.hexdigest _current_user.email}?s=30", alt: ''),
         content_tag(:span, _current_user.email),
-      ].compact.join.html_safe
+      ].map(&:presence).compact.join.html_safe
```
mshibuya, line 29:
```suggestion
      ].filter(&:present?).join.html_safe
```
will be a bit concise.

app/helpers/rails_admin/application_helper.rb
```diff
@@ -26,11 +26,11 @@ def edit_user_link
...
       if (edit_action = RailsAdmin::Config::Actions.find(:edit, controller: controller, abstract_model: abstract_model, object: _current_user)).try(:authorized?)
         link_to content, rails_admin.url_for(action: edit_action.action_name, model_name: abstract_model.to_param, id: _current_user.id, controller: 'rails_admin/main'), class: 'nav-link'
       else
-        content_tag :span, content
+        content.sub('<span>', '<span class="nav-link">').html_safe
```
mshibuya, line 33:
```suggestion
        content_tag :span, content, class: 'nav-link'
```
I would suggest this, because it's simpler and achieves the same style.

mshibuya:
Would you mind adding some tests for this fix? They should fit into somewhere around here:
https://github.com/railsadminteam/rails_admin/blob/master/spec/helpers/rails_admin/application_helper_spec.rb#L463-L506

app/helpers/rails_admin/application_helper.rb
thedanielhanke: this would cause nested spans, but would also fix the render issue.

thedanielhanke:
@mshibuya 

concerning the tests: sure, the only issue i had was getting rspec up and running when using sprockets. 

it failed at require sprockets/railstie - is this a known thing? will try webpacker later.

as rspec did not start and the current specs just dont really care for the generated html, my first attempt was just to quick-fix the rendering issue. would love to add some actual improvements..

mshibuya:
Ah you need to run rspec via appraisal, like this:

```bash
$  appraisal rails-7.0 rspec spec/helpers/rails_admin/application_helper_spec.rb
```

thedanielhanke:
thanks, working now.

question concerning `"#{request.ssl? ? 'https://secure' : 'http://www'}."`:

is there any real world requirement for the `http://www` part? why not always do https here?

thedanielhanke:
did some changes, please have a look. i paused, as i changed quite a bit - if you are not entirely against these kind of changes, i`ll continue.

kind of changes: imo. readability, maintainability, specs care a bit more about the actual outcome

mshibuya:
@thedanielhanke

> kind of changes: imo. readability, maintainability, specs care a bit more about the actual outcome.

Please don't do it now. You can propose these changes in a separate PR, but do not mix it with the issue fix.

mshibuya:
Merged in, thanks!

PR #3477, opened 2022-02-25
Fix layout of scope tabs

coorasse:
The layout was wrong. This PR adjusts the CSS classes according to Bootstrap5


mshibuya:
Specs are failing, could you fix them?
https://github.com/railsadminteam/rails_admin/runs/5332865440?check_suite_focus=true#step:7:71
https://github.com/railsadminteam/rails_admin/blob/master/spec/integration/actions/index_spec.rb#L989-L1069

mshibuya:
OK I can take care of the rest!

PR #3483, opened 2022-02-27
Enable frozen string literals across the project

jdufresne:
https://docs.rubocop.org/rubocop/cops_style.html#stylefrozenstringliteralcomment

This feature is designed to help transition from mutable string literals
to frozen string literals. The "frozen_string_literal: true" magic
comment was added to the top of all files to enable frozen string
literals. Frozen string literals may be default in future Ruby. Using
frozen string literals is good practice and easier to reason about.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -168,9 +168,6 @@ Style/Encoding:
 Style/EndlessMethod:
   Enabled: true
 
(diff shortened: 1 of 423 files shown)


codealchemy (approved):
👍🏼

Issue #3485, opened 2022-03-01
Have issue with data types Points (Postgrsql) 


mr3abd:
**Describe the bug**
Bug it's in Rails Admin Can't Access for data type points like 
<img width="1149" alt="Screen Shot 2022-03-01 at 11 24 26 PM" src="https://user-images.githubusercontent.com/17132856/156251331-79a2d4fe-1df7-4720-ab6a-efbe4f2a884e.png">

**Reproduction steps**
try to create db potgres 
and make any model like it 
`rails g model place location:points 
`
and will face the same error
I think about the solution add type points here in rails admin 
<img width="1396" alt="Screen Shot 2022-03-01 at 11 27 37 PM" src="https://user-images.githubusercontent.com/17132856/156251831-886467d5-c93b-478e-903a-765653e3ec16.png">

I'll try to make PR if have enough time

mr3abd:
I do that solution 
https://github.com/railsadminteam/rails_admin/pull/3486

PR #3486, opened 2022-03-01
for issue #3485 Add point filed

mr3abd:
That PR is just a simple fix for Postgres attribute type  (point )


mshibuya:
This doesn't add any value. If you want RailsAdmin to handle Postgres the POINT type, you need to add more implementation so it can be treated naturally by various RailsAdmin functionalities.
Also all changes must come with test codes.

mshibuya:
@mr3abd
> you need to add more implementation

Your change breaks with invalid inputs, does not allow updating the POINT field nor searching by the filter. If you want to add the field, you need to make it function sanely for end users.

Issue #3490, opened 2022-03-09
per model config can no longer be defined before associations


q3aiml:
**Describe the bug**
While testing an upgrade from rails_admin 2.x to `3.0.0.rc3`, I noticed that a number of our association fields were incorrectly using the `String` type. Previously on 2.x these fields correctly used an association type such as `BelongsToAssociation`.

I tracked this down to a behavior change when `rails_admin` config is defined in a model before the associations. In 2.x the execution of the `rails_admin` block is deferred and runs after the associations are defined. In 3.x the block runs immediately. As a result in 3.x any [field](https://github.com/railsadminteam/rails_admin/blob/b3ccb26ffb126870f6429a872d2a1532a4950725/lib/rails_admin/config/has_fields.rb#L8) declarations effectively disable the [field factories](https://github.com/railsadminteam/rails_admin/blob/b3ccb26ffb126870f6429a872d2a1532a4950725/lib%2Frails_admin%2Fconfig%2Ffields.rb#L48) and force a string type. The declarations no longer modify an existing field automatically created from the association and instead create the field.

I believe this may be caused by https://github.com/railsadminteam/rails_admin/commit/e4ae6698e52e56a1cefdf5c4097b29b8306f21e2.

**Reproduction steps**

Here is standalone example. Uncomment the relevant `gemfile` section for either 2.x or 3.x.
```
# frozen_string_literal: true

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  # >> uncomment this section to show rails_admin 3.x behavior
  # gem 'rails', '~> 7.0.0'
  # gem 'rails_admin', '~> 3.0.0.rc3', require: false
  # >> end

  # >> uncomment this section to show rails_admin 2.x behavior
  # gem 'rails', '~> 6.0.0'
  # gem 'rails_admin', '~> 2.2.1', require: false
  # gem 'sprockets', '<4'
  # >> end

  gem 'sqlite3'
end

require 'active_record/railtie'
require 'active_storage/engine'
require 'rails_admin'
require 'logger'

class TestApp < Rails::Application
  config.root = __dir__
  config.eager_load = false
  config.session_store :cookie_store, key: 'cookie_store_key'
  secrets.secret_key_base = 'secret_key_base'

  config.logger = Logger.new($stdout)
  Rails.logger  = config.logger
end

ENV['DATABASE_URL'] = 'sqlite3::memory:'

Rails.application.initialize!

RailsAdmin.config do |config|
[...]

Issue #3025, opened 2018-05-28
Active Storage: `method_missing': undefined method `has_one_attached'


cristiansimioni:
Hello guys, I'm facing a issue while using active storage with rails_admin.

What's happening is that I've model called Product which should have a image. I've installed active storage by running the following commands:

```
rails active_storage:install
db:migrate
```
After that, I've added a new attibute on my Model as:

```
has_one_attached :image
```

However, rails does not recognize it as a method:

> <class:Product>'
> /Users/username/.rvm/gems/ruby-2.5.1/gems/activerecord-5.2.0/lib/active_record/dynamic_matchers.rb:22:in `method_missing': undefined method `has_one_attached' for Product (call 'Product.connection' to establish a connection):Class (NoMethodError)

I've created a application from the scratch without rails_admin and it worked fine.

Note: I'm using rails **5.2.0** and ruby **2.5.1**

Any ideas?

dmilisic:
I think I've had the same issue. I was trying to configure model inside rails_admin initializer using `config.model Product do ... end`. Instead it should've been `config.model 'Product' do ... end`. The model name must be quoted.

cristiansimioni:
@dmilisic thank you very much! It saved me a lot of time. :)

But, any idea why it should be quoted?

dmilisic:
Initalizers are run before your application code runs. Models are interpreted only when needed (on a firs request perhaps). But if you reference a model in the initializer using a constant ie. `configure Product do`, the model has to be interpreted before the rest of your initializer code and other initalizers. Configuring your models using string `configure 'Product' do` works well with rails_admin.

It seems that `ActiveStorage` has its own initalizer which waits `ActiveRecord` to load [before it inserts its macros](https://github.com/rails/rails/blob/master/activestorage/lib/active_storage/engine.rb#L61). Perhaps `rails_admin.rb` initalizer is executed before `active_storage.attached` initalizer ?

bperlik:
If you don't use require "rails/all', you'll need to require "active_storage" in your application.rb file

tteurs:
Same issue here, is there any fix?

tteurs:
declaring this on model solved the problem for me
`class User < ApplicationRecord
  has_one_attached :avatar

  rails_admin do
    field :avatar, :active_storage
  end
end`

reference [https://github.com/sferik/rails_admin/issues/2990#issuecomment-378220331](url)

iarobinson:
[...]

PR #3492, opened 2022-03-10
restore 2.x config loading behavior

q3aiml:
Attempt to restore 2.x config loading behavior where 1) per-model config blocks
only evaluate after all the models are loaded and associations are defined
while 2) initializer config block evaluates immediately so that routes are
configured correctly.

One remaining change compared to 2.x is RailsAdmin::Config#model no longer
returns the model when given a block. With the removal of LazyModel in e4ae669
it is tricky to both defer initialization and return the model class.

Fixes https://github.com/railsadminteam/rails_admin/issues/3490

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -115,10 +115,10 @@ Metrics/CyclomaticComplexity:
 
 Metrics/MethodLength:
   CountComments: false
-  Max: 29 # TODO: Lower to 15
+  Max: 30 # TODO: Lower to 15
 
 Metrics/ModuleLength:
-  Max: 228 # TODO: Lower to 100
+  Max: 248 # TODO: Lower to 100
 
 Metrics/ParameterLists:
   Max: 8 # TODO: Lower to 4
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -118,7 +118,7 @@ Metrics/MethodLength:
   Max: 29 # TODO: Lower to 15
 
 Metrics/ModuleLength:
-  Max: 217 # TODO: Lower to 100
+  Max: 228 # TODO: Lower to 100
 
 Metrics/ParameterLists:
   Max: 8 # TODO: Lower to 4
--- a/lib/rails_admin.rb
+++ b/lib/rails_admin.rb
@@ -29,7 +29,9 @@ module RailsAdmin
     if entity
       RailsAdmin::Config.model(entity, &block)
     elsif block_given?
-      RailsAdmin::Config.apply(&block)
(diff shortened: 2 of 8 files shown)


mshibuya:
Thanks for trying! I like your idea, but

> 2) initializer config block evaluates immediately

there's a problem with this part. This will cause the issue with autoloading.

Sorry not covering it in the spec for now, I was unable to reproduce from the spec it since it happens during the initial loading stage. But you can reproduce it by using [the dummy app](https://github.com/railsadminteam/rails_admin/tree/master/spec/dummy_app), with changing [the initializer](https://github.com/railsadminteam/rails_admin/blob/master/spec/dummy_app/config/initializers/rails_admin.rb) as

```diff
 RailsAdmin.config do |c|
   c.asset_source = CI_ASSET
-  c.model 'Team' do
+  c.model Team do
     include_all_fields
     field :color, :color
   end
```

and starting the Rails app. You will get `uninitialized constant Team` error, since the Rails autoloader is not set up yet on the execution of the RailsAdmin initializer. But this notation `c.model <a model class>` is totally valid in the current DSL.
Hence my choice was to defer initialization to some extent...

Could you see if this is fixable?

q3aiml:
Thanks for the review! I didn't think about how that change was intentional with 3.x (unless I am mistaken I think the autoload error exists as well in 2.x when using `c.model <a model class>` in the initializer). After reviewing issues like https://github.com/railsadminteam/rails_admin/issues/3025 I can see how it could be less confusing to defer the initializer config too.

I made a first attempt at a fix. The idea is to have two separate lists of deferred config blocks: one for the initializer config (I called it "core"??) and one for the model config. It probably deserves a little cleanup but I want to get it out for your thoughts before getting too far along. Currently it does change some methods that could be considered public api with the release of 3.0.0 so that might deserve some rework as well.

jeromedalbert:
Thanks for making this PR, I am having class loading issues on staging/production when I use rails_admin 3.0 along with the [makara](https://github.com/instacart/makara) gem, and your PR fixes it by restoring the old 2.x behavior. I hope it gets merged.

mshibuya:
Sorry for not getting back for a long time, I needed some time to consider all implications of the initialization process.
And I came to realize that the change proposed in this PR is not compatible with the Importmap support #3488, which requires `asset_source` to be set earlier than `after_initialize`.

So I'm proposing another way to address this as #3541. It would be great if you could take a look.

q3aiml:
Thanks for circling back with the update! It's definitely a tricky problem. And good catch on the importmap incompatibility.

At first look #3541 makes sense to me. I'll give it a try shortly.

PR #3455, opened 2022-01-04
Upgrade Bootstrap to 5.1.3 with migrating datetimepicker to Flatpickr

mshibuya:
image

Closes #3083

# Reason for upgrading Bootstrap to 5

- Keeping our dependency up to date is necessary for our project to continue. (#3083)

# Reason for replacing Bootstrap Datetimepicker with Flatpickr

- Bootstrap Datetimepicker depends on Bootstrap 3. Since we're upgrading Bootstrap, it also needs to be updated.
- Bootstrap Datetimepicker also depends on momentjs, which is a project stated as in [maintenance mode](https://momentjs.com/docs/).

## Considered options

1. Keep using [Bootstrap Datetimepicker](https://getdatepicker.com/4/)
    - Applying a minor patch may enable us to keep using this.
    - But this version is no longer maintained, with the project name changed to Tempus Dominus below.
2. Migrate to [Tempus Dominus v5](https://getdatepicker.com/5-4/)
    - This is a stable version, but it doesn't come with Bootstrap 5 integration (only 3 and 4).
    - Also this depends on jQuery, which contradicts with our general direction of [removing it](https://github.com/railsadminteam/rails_admin/issues/2893).
3. Migrate to [Tempus Dominus v6](https://github.com/Eonasdan/tempus-dominus)
    - Having no jQuery dependency is good, but it's still in alpha.
4. Migrate to [Flatpickr](https://flatpickr.js.org/) with [Day.js](https://day.js.org/)
    - Flatpickr having no jQuery dependency is good.
    - Also Day.js can provide functionality equivalent to momentjs, and it can easily be integrated with Flatpickr.
    - But I found working with Day.js can take a bit effort, especially with i18n things.
5. **Migrate to Flatpickr without a date library**
    - This is the option I decided to use, it is simple enough while covering majority of usecases.
    - But this will require users to [rewrite](https://github.com/railsadminteam/rails_admin/pull/3455#discussion_r779587282) their `momentjs_format` to Flatpickr's one if they use it.
6. Migrate to other datepicker libraries
    - I took a glance at many other libraries, but many of them
      - have jQuery dependency, or dependency for other big libraries we don't want.
      - have appearances which I don't like so much (just my personal preference...!).
      - are not so actively maintained.
    - If you're aware of other potential candidates please let me know!

PR #3493, opened 2022-03-10
restore full width for content

q3aiml:
During the bootstrap 5 update in 6878670 (#3455) this class was changed from
`content` to `container`. The `container` class, unlike `content`, has a max
width of 1320px. Restore prior behavior of using the full width by using
`container-fluid` to avoid wasted whitespace. This is especially useful on
list pages with a large number of columns.

<details>
<summary>Before</summary>

before non-full width image

</details>

<details>
<summary>After</summary>

after full width image

</details>

<details>
<summary>Bootstrap 3</summary>

image

</details>


mshibuya:
Ah I didn't notice, thanks!

PR #3495, opened 2022-03-12
Fixed problem with sprockets being the default during installation.

yubele:
https://github.com/railsadminteam/rails_admin/pull/3488

This solves the problem that occurs during `rails_admin:install` in the PR above.


mshibuya:
Thanks for the PR, but this fix is not good because with it gives users no choice of the asset delivery method they want to use. One may want choose to use :sprockets while having `config/importmap.rb` in the app.
The issue should be fixed by 705fd32fcea77978ed867d66696464da9b741a11 anyway, so closing this.

PR #3504, opened 2022-04-04
Add whitespace after navigation icon

fuegas:
As discussed by mshibuya and codealchemy in [upgrade to Bootstrap v5](https://github.com/railsadminteam/rails_admin/pull/3455/files#r780533297) the rendering of icons should not be fixed with CSS but adding a whitespace between the icon and the text.

This change applies that solution to the icons in the sidebar navigation as well.

### DIFF
--- a/app/helpers/rails_admin/application_helper.rb
+++ b/app/helpers/rails_admin/application_helper.rb
@@ -94,7 +94,7 @@ module RailsAdmin
       actions(:root).select(&:show_in_sidebar).group_by(&:sidebar_label).collect do |label, nodes|
         li_stack = nodes.map do |node|
           url = rails_admin.url_for(action: node.action_name, controller: 'rails_admin/main')
-          nav_icon = node.link_icon ? %(<i class="#{node.link_icon}"></i> ).html_safe : ''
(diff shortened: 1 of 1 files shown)


app/helpers/rails_admin/application_helper.rb
```diff
@@ -94,7 +94,7 @@ def root_navigation
       actions(:root).select(&:show_in_sidebar).group_by(&:sidebar_label).collect do |label, nodes|
         li_stack = nodes.map do |node|
           url = rails_admin.url_for(action: node.action_name, controller: 'rails_admin/main')
-          nav_icon = node.link_icon ? %(<i class="#{node.link_icon}"></i>).html_safe : ''
+          nav_icon = node.link_icon ? %(<i class="#{node.link_icon}"></i> ).html_safe : ''
```
mshibuya, line 97: We don't need this, since we have `link_to nav_icon + " " + wording_for...` just below this.

fuegas:
@mshibuya any update on the status of this PR? 😄

mshibuya:
Sorry for delay, thanks!

PR #3505, opened 2022-04-06
✨ Listing: ability to add a default_scope

RocKhalil:
In some cases, we want to list relationships, so it's better to include them ( usage of `includes` )
This feature will allow you to use a `default_scope` to the list to allow the users to use whichever scope they want in their list.

This can override the use of `scopes` when there's only ONE scope for example

### DIFF
--- a/lib/rails_admin/config/actions/index.rb
+++ b/lib/rails_admin/config/actions/index.rb
@@ -33,14 +33,22 @@ module RailsAdmin
         register_instance_option :controller do
           proc do
             @objects ||= list_entries
+            scope_to_use = :all
 
+            # If the user configured a default_scope, we use it
+            scope_to_use = @model_config.list.default_scope.to_sym if @model_config.list.default_scope.present?
+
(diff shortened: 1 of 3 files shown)


RocKhalil:
Or we can change this line in `app/views/rails_admin/main/index.html.erb:95`  to:
```
<% if (@model_config.list.scopes || []).count > 1 %>
```

mshibuya:
Recently the `scope` option was introduced to the model config. Isn't this what you want?
https://github.com/railsadminteam/rails_admin/commit/8d905f9e2f1102e8addf324b496720e3fd47bf1d#diff-2f13d2a41af18228eec779b8819dbc69a0d015d101660c9d0d3c12cee6fece37R970-R974

RocKhalil:
@mshibuya this is weird, never saw this feature, maybe because there's no documentation about it ?

mshibuya:
Honestly yes, I'm just writing it 😓

PR #3506, opened 2022-04-06
Fix orderable multiselect buttons

aaronlifton:
The Up/Down buttons for a "sortable" multiselect have labels, while the Left/Right buttons do not. These labels are cut off by the multiselect box. This PR removes the labels and uses the regional text as a `title` on the button instead.

**Before**
orderable bug example 1

**After**
orderable fix

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.swp
 .bundle
 .idea/
+.vscode/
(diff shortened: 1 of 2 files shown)


mshibuya:
Great catch, thanks!

PR #3509, opened 2022-04-08
Fix default sort by behavior when list.sort_by points to a field with a table reference for :sortable

aaronlifton:
On a list page, when a sort param is not specified, the current version of rails admin defaults to appending the `list.sort_by` to the model table. This causes `list.sort_by` to always  be treated as a string when a sort param is not present. This changes the behavior to lookup the field specified in `list.sort_by` rather than treating it as a string.

This also updates the dummy gemfile to point to devise's main branch, as they changed the `master` branch to the name `main`.

For example, if we have the following:
```
RailsAdmin.config('Player') do
    base do
      field :name do
        sortable :'teams.name'
      end
    end

    list do
      sort_by :name
    end
 end
```

the query would be generated with a sort by `players.name`, when we actually want to sort by `teams.name`.

### DIFF
--- a/app/controllers/rails_admin/main_controller.rb
+++ b/app/controllers/rails_admin/main_controller.rb
@@ -59,13 +59,15 @@ module RailsAdmin
       params[:return_to].presence && params[:return_to].include?(request.host) && (params[:return_to] != request.fullpath) ? params[:return_to] : index_path
     end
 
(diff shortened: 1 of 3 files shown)


mshibuya:
Agreed, thanks!

PR #3511, opened 2022-04-19
Update hover / active behavior for side navigation links

codealchemy:
We noticed that the sidenav active / hover behavior from RailsAdmin 2.x did not carry over to RailsAdmin 3.0. This adds `hover` behavior to the side nav, and the `active` class to the side nav element that corresponds to the current model context - both of which can be overridden in user themes to maintain the current behavior. 

side-nav-check

📝 I'm having trouble getting the 2.x branch running to demo the old behavior, but it can be seen in the [demo app](http://rails-admin-tb.herokuapp.com/admin/draft) linked from the README

### DIFF
--- a/src/rails_admin/styles/base/theming.scss
+++ b/src/rails_admin/styles/base/theming.scss
@@ -42,6 +42,16 @@ body.rails_admin {
     > li > a {
       padding: ($navbar-padding-y / 2) ($grid-gutter-width / 2);
 
+      &.active {
+        background-color: map-get($theme-colors, primary);
+        color: color-contrast(map-get($theme-colors, primary));
+      }
+
+      &:hover {
+        background-color: map-get($theme-colors, primary);
+        color: color-contrast(map-get($theme-colors, primary));
+      }
+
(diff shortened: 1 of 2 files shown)


src/rails_admin/styles/base/theming.scss
```diff
@@ -42,6 +42,16 @@ body.rails_admin {
...
+      &.active {
+        background-color: map-get($theme-colors, primary);
+        color: color-contrast(map-get($theme-colors, primary));
+      }
+
+      &:hover {
+        background-color: map-get($theme-colors, primary);
+        color: color-contrast(map-get($theme-colors, primary));
+      }
+
```
codealchemy, line 54: The styles here can be removed if it'd be preferable to keep with the current behavior, https://github.com/railsadminteam/rails_admin/pull/3511/commits/410654d76e7de5b5c560ca51ef7a92484b1ff75f is the main change needed to allow targeting the (active sidenav) element.

mshibuya:
👍

PR #3515, opened 2022-05-02
Use status instead of statusText

fuegas:
As of HTTP/2 the field statusText can be empty ( See: https://github.com/jquery/api.jquery.com/commit/be6de733d57d67634ef07dcfc38163d18945c0e4 ). Thus the field statusText is no longer relyable to base logic on. The field status of the XMLHttpRequest remains intact, so we can base the logic on that field.

We encountered this problem during testing of RailsAdmin 3.0 on local development environments versus acceptance environments. On the acceptance environments the success events wouldn't be fired anymore because the `statusText` was empty and thus the check `status == "OK"` always returned `false`.


mshibuya:
Looks good, thanks!

PR #3527, opened 2022-06-05
Support Composite Primary Keys gem

mshibuya:
Makes RailsAdmin compatible with [Composite Primary Keys](https://github.com/composite-primary-keys/composite_primary_keys). The compatibility logic is implemented within [a newly-created adapter](https://github.com/railsadminteam/rails_admin/compare/composite-primary-keys?expand=1#diff-90131f22f7e53326a1ed29f457896db3f08965cc1d4b79d5c9274ec670863328) as much as possible, so it doesn't affect existing implementation much.


mshibuya:
Let me merge this in!

PR #3531, opened 2022-06-17
Generate correct user edit url

### LINKED ISSUE #3514: User link (top right hand side) is incorrectly formed as user?action_name=edit&id=<user id>
**Describe the bug**
When developing within rails_admin (standing on the shoulders of giants) I often click on the user link to go into my own user record to check out features. In rails_admin 3.0, this link is incorrect.

**Reproduction steps**
Click on the User Link on any page heading, and the address bar shows the following which isn't valid..
> admin/user?action_name=edit&id=user_id
eg
> admin/user?action_name=edit&id=5

**Expected behavior**
The current user record should open called from the link..
> admin/user/user_id/edit
eg
> admin/user/5/edit

**Additional context**

- `rails` version: 6.1.5
- `rails_admin` version: 3.0
- `rails_admin` npm package version: using sprockets

**Possible fix**
I think the error is in `ApplicationHelper`
```
 def edit_user_link
....
        edit_url = rails_admin.url_for(
line 32:  action_name: edit_action.action_name,
          model_name: abstract_model.to_param,
          controller: 'rails_admin/main',
          id: _current_user.id,
        )
```
`Line 32` should be..
> action ~~name~~: edit_action.action_name,`

**Work around**
I've implemented a workaround in javascript run on each page by putting into `custom/ui.js`
```
    var userRootChildren = $('.edit_user_root_link').children(),
      href = userRootChildren.attr('href');
    userRootChildren.attr('href',
      href.split('?')[0] + '/' + href.split('id=')[1] + '/edit');
```

fuegas:
The current generated URL looks like `/user?action_name=edit&id=42`. However, if you use an authorization gem like cancancan this url requires the capability `:index`. If you only allow a user to edit itself, for example `can :edit, ::User, id: user.id`, the current URL would cause an Error to be thrown ( CanCan::AccessDenied ).

When we change action_name to action, the generated URL becomes `/user/42/edit` which matches the capability `can :edit, ::User, id: user.id`.

Fixes #3514.

### DIFF
--- a/app/helpers/rails_admin/application_helper.rb
+++ b/app/helpers/rails_admin/application_helper.rb
@@ -31,7 +31,7 @@ module RailsAdmin
 
       if authorized
         edit_url = rails_admin.url_for(
-          action_name: edit_action.action_name,
+          action: edit_action.action_name,
           model_name: abstract_model.to_param,
           controller: 'rails_admin/main',
           id: _current_user.id,


app/helpers/rails_admin/application_helper.rb
```diff
@@ -31,7 +31,7 @@ def edit_user_link
 
       if authorized
         edit_url = rails_admin.url_for(
-          action_name: edit_action.action_name,
```
codealchemy, line 34: It looks like this was originally `action`, and changed to `action_name` in https://github.com/railsadminteam/rails_admin/pull/3475 - @thedanielhanke was that needed for Pundit? Or maybe made in error?

codealchemy (commented):
One question to clarify (since this was recently changed), otherwise LGTM

app/helpers/rails_admin/application_helper.rb
mshibuya: I believe it was by mistake, and I overlooked it too.

mshibuya:
Thanks for the fix!

PR #3539, opened 2022-07-08
Adds focus: false to newly created modal dialogs

### LINKED ISSUE #3538: CKEditor modals don't work inside bootstrap modals
**Describe the bug**
When adding a link, or any other action in CK editor that generates a modal, text input fields no longer are selectable.

**Reproduction steps**
Have a form with a ckeditor field generate inside a modal (has_many relationships will work) and try and add a link.

**Expected behavior**
You can add the link and edit the link text.

**Additional context**

- `rails` version: 7.0.3
- `rails_admin` version: 3.1.0.beta (and 3.0.0)

The fix is to update one line (PR incoming) to tell modals in bootstrap not to steal focus all the time.

tosbourn:
Without forcing focus false, the modal will not allow other modals to have focus for their inputs. This breaks tools like CKEditor.

From local testing, there are no negative side effects with not forcing focus on open.

Closes #3538

### DIFF
--- a/src/rails_admin/remote-form.js
+++ b/src/rails_admin/remote-form.js
@@ -183,6 +183,7 @@ import * as bootstrap from "bootstrap";
         new bootstrap.Modal(widget.dialog[0], {
           keyboard: true,
           backdrop: true,
+          focus: false,
           show: true,
         }).show();
       }


codealchemy (approved):
👍

Issue #3534, opened 2022-07-01
Unable to precompile assets when updating to `3.0.0.beta`


codealchemy:
**Describe the bug**

Attempting to run `bundle exec rake assets:precompile` raises an exception

**Reproduction steps**

Update gem in a project using `:sprockets`, attempt to precompile

**Expected behavior**

Assets to be compiled as before

**Additional context**

- `rails` version: `6.0.5`
- `rails_admin` version: `3.0.0.beta`
- `rails_admin` npm package version: N/A
- full stack trace (if there's an exception)

(reduced this down to the essential bit of the error as we have internal middlewares in addition to those from Rails)
```
FrozenError:
  can't modify frozen Array: [... middleware classes ]
/project_dir/vendor/bundle/ruby/2.7.0/gems/actionpack-6.0.5/lib/action_dispatch/middleware/stack.rb:126:in `push'
/project_dir/vendor/bundle/ruby/2.7.0/gems/actionpack-6.0.5/lib/action_dispatch/middleware/stack.rb:126:in `use'
/project_dir/vendor/bundle/ruby/2.7.0/gems/<internal gem>:in `block in <class:Engine>'
/project_dir/vendor/bundle/ruby/2.7.0/gems/railties-6.0.5/lib/rails/initializable.rb:32:in `instance_exec'
/project_dir/vendor/bundle/ruby/2.7.0/gems/railties-6.0.5/lib/rails/initializable.rb:32:in `run'
/project_dir/vendor/bundle/ruby/2.7.0/gems/railties-6.0.5/lib/rails/initializable.rb:61:in `block in run_initializers'
/project_dir/vendor/bundle/ruby/2.7.0/gems/railties-6.0.5/lib/rails/initializable.rb:60:in `run_initializers'
/project_dir/vendor/bundle/ruby/2.7.0/gems/railties-6.0.5/lib/rails/application.rb:363:in `initialize!'
```

This appears to be related to changes to initialization from https://github.com/railsadminteam/rails_admin/commit/e59a9353e742b0f6e6b152d1b1f9e7cafad3ee54

**Workaround**

Keep only the precompile initializer during precompilation

```
# application.rb

if ENV.fetch('PRECOMPILE', false)
  RailsAdmin::Engine.initializers.keep_if { |i| i.name == 'RailsAdmin precompile hook' }
end
```

codealchemy:
[...]

Issue #3533, opened 2022-06-28
Reloading Routes before all cause Devise initializers not to be loaded dependent to the order of gem requires


jklimke:
Hi,

i noticed that reloading routes after causes to draw devise routes before all devise initializers have been executed. It possibly  depends on the order of the gems within the gemfile. When devise gems are required  before rails admin gems, everything runs smooth, if they are after rails_admin the devise initializers are not yet executed and though the devise routes cannot be drawn (e.g. not executing this initializer causes route drawing to fail https://github.com/heartcombo/devise/blob/6d32d2447cc0f3739d9732246b5a5bde98d9e032/lib/devise/rails.rb#L36)

This is the line that causes the route reload. Maybe this could be moved into a callback after finishing reload. The quick fix for me was to include devise before rails admin that causes devise initializers to run prior to rails_admin. maybe as the simplest "fix" this could be emphazized in the docs.

https://github.com/railsadminteam/rails_admin/blob/a2e366e8839c9046b9f9fde219875a05bc1a66bb/lib/rails_admin/engine.rb#L42

Issue #3501, opened 2022-03-28
Relation does not exist


cristofer:
**Describe the bug**
After upgrading to rails-admin 3.0.0, and running `rails g rails_admin:install`, in our CI we see that one of the relations does not exist

**Reproduction steps**
- Upgrade to latest version (3.0.0)
- Run `rails g rails_admin:install`
- Apply what the install process recommended, on this case was to add the gem `sassc-rails` and to add `config.asset_source = :sprockets` on `config/initializers/rails_admin.rb`
- Push the changes to Circle CI
- Step `rails db:create` fails with:

 ```
DEPRECATION WARNING: The total_columns_width configuration option is deprecated and has no effect. (called from block in <main> at /home/circleci/repo/config/initializers/rails_admin.rb:13)
rake aborted!
ActiveRecord::StatementInvalid: PG::UndefinedTable: ERROR:  relation "TABLE" does not exist
LINE 8:  WHERE a.attrelid = '"TABLE"'::regclass
```  

**Expected behavior**
`rails db:create` should not fail, the table is defined in our `schema`, and of course the model exists

**Additional context**

- `rails` version: 6.1.5
- `rails_admin` version: 3.0.0 (from 2.2.1)
- `rails_admin` npm package version:
- full stack trace (if there's an exception)

```
ActiveRecord::StatementInvalid: PG::UndefinedTable: ERROR:  relation "lalas" does not exist
LINE 8:  WHERE a.attrelid = '"lalas"'::regclass
                            ^
/home/circleci/repo/vendor/bundle/ruby/3.1.0/gems/activerecord-6.1.5/lib/active_record/connection_adapters/postgresql/database_statements.rb:19:in `exec'
/home/circleci/repo/vendor/bundle/ruby/3.1.0/gems/activerecord-6.1.5/lib/active_record/connection_adapters/postgresql/database_statements.rb:19:in `block (2 levels) in query'
/home/circleci/repo/vendor/bundle/ruby/3.1.0/gems/activesupport-6.1.5/lib/active_support/dependencies/interlock.rb:48:in `block in permit_concurrent_loads'
/home/circleci/repo/vendor/bundle/ruby/3.1.0/gems/activesupport-6.1.5/lib/active_support/concurrency/share_lock.rb:187:in `yield_shares'
[...]

PR #3541, opened 2022-07-10
Load constants lazily on RailsAdmin initialization

mshibuya:
This is a counter-proposal for #3492, and aims to fix issues #3490, #3501, #3533, and #3534.

My apologies for those issues. This is because of the complexity in RailsAdmin (and Rails) initialization process, which is hard to get it right. Let me explain the story...

RailsAdmin initializer (`config/initializers/rails_admin.rb`) is loaded during the Rails app's startup process. But as for the actual configuration (= execution of the DSL itself, the code passed in the `RailsAdmin.config do |config| ... end` block), there're many constraints that all need to be fulfilled for a successful initialization.

- At this point, basically dependent gems are already loaded but the application codes (e.g. files under `app/` directory) are not loaded yet.
- But historically RailsAdmin [allows model constants to be passed as the `config.model` argument](https://github.com/railsadminteam/rails_admin/pull/3492#issuecomment-1064915362). This led to issues like #3025 in pre-Zeitwerk age, and with Zeitwerk it's not possible because its autoload feature is not enabled at this moment.

So the attempt to address this was made in [e4ae669](https://github.com/railsadminteam/rails_admin/commit/e4ae6698e52e56a1cefdf5c4097b29b8306f21e2). The idea was basically to delay the execution of the configuration block, after model classes are ready to be autoloaded. But it brought different problems:

- Devise [performs `reload_routes!` on `before_eager_load`](https://github.com/heartcombo/devise/blob/8593801130f2df94a50863b5db535c272b00efe1/lib/devise/rails.rb#L17). That means RailsAdmin actions ([which are used to construct RailsAdmin routes](https://github.com/railsadminteam/rails_admin/blob/master/config/routes.rb)) need to be set up ahead of that, unless RailsAdmin routes fail to be reflected. To work aronud this issue, `reload_routes!` was [also added](https://github.com/railsadminteam/rails_admin/commit/08f50aabc1922aeb289ced4ccbe9f983bc1aaa89) in the RailsAdmin initialization process. And that solution isn't perfect, resulting in #3533.
- Upon the introduction of [the Importmap support](https://github.com/railsadminteam/rails_admin/pull/3488), [RailsAdmin's asset configuration](https://github.com/railsadminteam/rails_admin/blob/e59a9353e742b0f6e6b152d1b1f9e7cafad3ee54/lib/rails_admin/engine.rb#L45) must happen earlier than `after_initialize` (at that point, it's too late to modify values in `app.config.assets`). So [the configuration execution](https://github.com/railsadminteam/rails_admin/blob/e59a9353e742b0f6e6b152d1b1f9e7cafad3ee54/lib/rails_admin/engine.rb#L38) was brought earlier, but that didn't go well and brought #3534.

Thus, it's like 2 contradicting forces pulling the initialization timing both forward ant backward. That's why I have been struggling in solving this issue without success...

We need a new way to overcome this problem, and here I have come up with one. That is to intercept `const_missing` to eliminate the necessity for loading actual classes, while delaying the model initialization by utilizing `LazyModel`, which is the proven method from pre RailsAdmin 3.x era.
At the first sight overriding `Object.const_missing` seemed to have side effects, but from what I tested it looks working well. So I want you guys to try this in your setup, and give feedback on whether this is good idea or not.

Thank you for reading this through, and hoping to hearing back!

@codealchemy @jdufresne @q3aiml (or anyone who is interested in this)


spec/dummy_app/config/initializers/rails_admin.rb
```diff
@@ -2,8 +2,8 @@
 
 RailsAdmin.config do |c|
   c.asset_source = CI_ASSET
-  c.model 'Team' do
```
codealchemy, line 5: I think this was where I got tripped up in my config (using model constants in the initializer), leading to https://github.com/railsadminteam/rails_admin/issues/3534

It feels problematic to continue to allow this (and makes sense to me that one shouldn't reference model _constants_ in an initializer, https://github.com/railsadminteam/rails_admin/issues/3542 highlights another possibly related issue). If need be, the config block for a model can already live in the model itself (which helps keep the initializer more manageable in large applications).

I'd be in favor of removing support for using model constants in the initializer entirely (given that using the model name as a String continues to work as expected), particularly if that can help avoid needing to override `const_missing`

codealchemy (commented):
I can confirm that this change addresses the issues I was running into in https://github.com/railsadminteam/rails_admin/issues/3534

Some thoughts re: allowing model constants to be used in the initializer noted above.

spec/dummy_app/config/initializers/rails_admin.rb
codealchemy: Got it, I think it works well given the cases needed to be handled here - just always cautious with overrides

codealchemy (approved):

mshibuya:
Merged. Thank you for reviewing!

q3aiml:
This is working well in my testing and resolves https://github.com/railsadminteam/rails_admin/issues/3490. Thank you for figuring this out!

PR #3544, opened 2022-07-15
Fix dart sass 2.0 division deprecations

darrenterhune:
The `src/rails_admin/styles/base/theming.scss` used division calculations which causes deprecation warnings with dart sass >= 2.

The fontawesome-free version also has the same problem.

Using calc() around the division math fixes the issue in theming.scss
Upgrading fontawesome-free to the latest version fixes their issues

https://github.com/sass/sass/issues/2565
https://github.com/sass/dart-sass/issues/663

For rails projects using sassc this is a problem. The app I currently work on is on most recent gems and packages for everything. We do a really good job at keeping everything up to date. Unfortunately these warnings have now turned into errors being raised in `webpacker-dev-server`.

I tested these changes by publishing my own npm package while pointing our `Gemfile` to my forked version.

If we could get a minor release with this and any other minor fixes that may be in that would be really appreciated 😄


codealchemy (approved):
👍 (`calc` has been available in `dart-sass` [since 1.40.0](https://github.com/sass/dart-sass/releases/tag/1.40.0))

mshibuya:
Nice fix!
There are also vendorized [css](https://github.com/railsadminteam/rails_admin/blob/master/vendor/assets/stylesheets/rails_admin/font-awesome.scss) and [fonts](https://github.com/railsadminteam/rails_admin/tree/master/vendor/assets/fonts/rails_admin)), to be used by Sprockets-only applications. Could you update them as well?

darrenterhune:
@mshibuya yeah I can update those today 😄

wartron:
Thanks for fixing this.

mshibuya:
Thank you 🌟

PR #3548, opened 2022-08-10
Add primary_key to the end of order

gogutan:
When I use rails_admin with PostgreSQL(latest 14.4), it is unstable to order by columns which are not unique.
Therefore, some records appear in several pages while some don't in any pages.
I would like to solve this problem by adding its primary_key to the end of order.

### Before

Player___Dummy_App_Admin
Player___Dummy_App_Admin

### After

Player___Dummy_App_Admin
Player___Dummy_App_Admin

### DIFF
--- a/lib/rails_admin/adapters/active_record.rb
+++ b/lib/rails_admin/adapters/active_record.rb
@@ -117,7 +117,7 @@ module RailsAdmin
         direction = options[:sort_reverse] ? :asc : :desc
         case options[:sort]
         when String, Symbol
-          scope.reorder("#{options[:sort]} #{direction}")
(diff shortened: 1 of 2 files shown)


mshibuya:
I'm not a fan of this change. Sorting data is a heavy operation for RDBMSs, and doing this will always add an extra sorting, including cases that have stable order.
A better option will be supporting ordering by multiple columns, so a user can choose to add an extra sorting when needed.

jdufresne:
Thanks, I'm convinced. I like your better option proposal!

gogutan:
I'm sorry for leaving some failures for a while, and thank you for your alternative solution!
Although I close this PR, hopefully I will try to make other PRs in future!

PR #3555, opened 2022-10-04
Resolved a problem with embedded mongoid documents and boolean selectors

rpalackas:
Previous implementation was creating a label with an incorrect for attribute when used on an embedded mongoid document. It would create <label for=parent_class[embedded_class_attributes]_field_name_1 ... when the id of the input field was parent_class_embedded_class_attributes_field_name_1. Because of this the selectors did nothing when clicked since the label references an id that does not exist. This change to using the rails label helper should safely create the label in all cases.


mshibuya:
Could you write a spec for this fix?
https://github.com/railsadminteam/rails_admin/blob/master/spec/integration/fields/boolean_spec.rb
will be the good place to add it.

sasselin:
This fix work. Have the bug with 3.1.2 too

rpalackas:
@mshibuya Are there any known random failures in the specs? According to the CI I've got failures on rails_6.1 and 7 but running those locally passes for me
bundle exec appraisal rails-6.1 rspec
...
1099 examples, 0 failures, 1 pending

bundle exec appraisal rails-7.0 rspec
...
1097 examples, 0 failures, 1 pending

rpalackas:
Well reducing the js scope resolved some of my failures. Not sure what would cause the remaining one though any feedback here would be appreciated

mshibuya:
I can handle the rest. Thank you!

PR #3559, opened 2022-10-27
Tweak `select` form fields class for filters

codealchemy:
Bootstrap 5 uses `form-select` (https://getbootstrap.com/docs/5.0/forms/select/) to help indicate the field is a dropdown.

#### Before

Capture d’écran 2022-10-27 à 13 54 24

#### After

Capture d’écran 2022-10-27 à 13 54 35


mshibuya:
It looks better, thanks! 👍

PR #3561, opened 2022-11-05
Simplify uses of defined?

jdufresne:
If an instance variable isn't defined it will always return nil.

### DIFF
--- a/lib/rails_admin/config/model.rb
+++ b/lib/rails_admin/config/model.rb
@@ -48,9 +48,7 @@ module RailsAdmin
       end
 
       def excluded?
-        return @excluded if defined?(@excluded)
-
-        @excluded = !RailsAdmin::AbstractModel.all.collect(&:model_name).include?(abstract_model.try(:model_name))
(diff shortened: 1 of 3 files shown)


lib/rails_admin/config/model.rb
```diff
@@ -48,9 +48,7 @@ def initialize(entity)
       end
 
       def excluded?
-        return @excluded if defined?(@excluded)
-
-        @excluded = !RailsAdmin::AbstractModel.all.collect(&:model_name).include?(abstract_model.try(:model_name))
+        @excluded ||= !RailsAdmin::AbstractModel.all.collect(&:model_name).include?(abstract_model.try(:model_name))
```
tachyons, line 51: Is the behaviour change here intended?  This change removes the benefit of memoisation when the result is falsely.

Ref: http://gavinmiller.io/2013/advanced-memoization-in-ruby/

jdufresne: Good catch. No that is not intended. If you put up a PR I'll approve, otherwise I'll try to get to it later today.

jdufresne: Restored in #3587. Thanks for reporting.

PR #3571, opened 2022-11-22
Define jQuery object in separate file to support esbuild

wayc:
### Background
During our migration from webpacker to esbuild, our apps using `rails_admin` began to fail during builds, along with Chrome showing a `jQuery is not defined` error in the Console:

image

We traced the root cause to the order of these actions in `src/rails_admin/base.js`:
- [Line 3](https://github.com/railsadminteam/rails_admin/blob/9997c1095066aaac39afb27bf8de705cf6ccb1ef/src/rails_admin/base.js#L3): import `jquery`
- [Lines 8-18](https://github.com/railsadminteam/rails_admin/blob/9997c1095066aaac39afb27bf8de705cf6ccb1ef/src/rails_admin/base.js#L8-L18): import `jquery-ui`
- [Line 30](https://github.com/railsadminteam/rails_admin/blob/9997c1095066aaac39afb27bf8de705cf6ccb1ef/src/rails_admin/base.js#L30): `window.$ = window.jQuery = jQuery;`

The `jquery-ui` modules are imported before `window.$` is defined, causing the above issues.

To add support for esbuild and to avoid issues with async and hoisting, we suggest importing and initialising `jQuery` together.

### Changes
- Move the jQuery initialiser into a separate file to import and define jQuery together

### DIFF
--- a/lib/rails_admin/config/configurable.rb
+++ b/lib/rails_admin/config/configurable.rb
@@ -57,7 +57,7 @@ module RailsAdmin
 
           # If it's a boolean create an alias for it and remove question mark
           if option_name.end_with?('?')
-            scope.send(:define_method, "#{(+option_name).chop!}?") do
+            scope.send(:define_method, "#{option_name.chop!}?") do
               send(option_name)
             end
           end
--- a/lib/rails_admin/config/configurable.rb
+++ b/lib/rails_admin/config/configurable.rb
@@ -57,7 +57,7 @@ module RailsAdmin
 
           # If it's a boolean create an alias for it and remove question mark
           if option_name.end_with?('?')
-            scope.send(:define_method, "#{option_name.chop!}?") do
+            scope.send(:define_method, "#{(+option_name).chop!}?") do
               send(option_name)
             end
           end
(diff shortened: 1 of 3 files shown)


rjaus:
Nice work.  Are these the only changes you made to get rails_admin running with esbuild?  I'm struggling to get it to work.

wayc:
> Nice work. Are these the only changes you made to get rails_admin running with esbuild? I'm struggling to get it to work.

@rjaus Hi there 👋 

Yes, the changes in this PR allowed esbuild to successfully build. Our migration from webpacker to esbuild took some effort. If you're in a similar situation, you can try creating a new separate Rails project on esbuild: `rails new demo-rails-with-react-frontend -j esbuild` and implementing `rails_admin`. Then you can diff between your new and existing projects to identify areas to change.

Looks like there are a few test suites failing here—I'll sort them out soon.

wayc:
@mshibuya @pcai 

Thank you so much for resolving the failing JRuby tests in https://github.com/railsadminteam/rails_admin/pull/3574

I've marked this PR (on jQuery initialisation) as ready to review and would love to hear your thoughts.

mshibuya:
Confirmed to work, thanks!

PR #3575, opened 2022-11-26
Preserve order when generating CSV

### LINKED ISSUE #2580: CSV export doesn't keep order
I just came across the need to sort and filter data and export it as a CSV file. While filtering and exporting works just fine the order is reset to the default database order. It doesn't keep the order from the list view for exported data. Is this the expected behavior and if so, could we change this?

I hope I didn't miss something. Thanks in advance!

pcai:
Resolves #2580

Generate CSVs using kaminari pagination when available (e.g., when `@objects` is an [ActiveRecord_Relation](https://api.rubyonrails.org/v6.1.0/classes/ActiveRecord/Relation.html)) instead of using the native rails [find_each](https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each).

This allows us to preserve the activerecord ORDER BY clause defined when browsing records before attempting to export them; `find_each` overrides the ORDER BY to use the primary key id. 

One potential downside to this is that it will make certain large exports much less efficient, as paginating large tables ordered by a column without the correct database indexes will run into [algorithmic problems](https://wiki.c2.com/?ShlemielThePainter).

Issue reproduction:
* Navigate to a records index page and change the sort ordering (e.g., sort by "Name descending"):
  image
* Click "export found _____"
* generate the CSV

EXPECTED:
* the generated CSV has the sort ordering from the previous step

ACTUAL:
* the generated CSV is in primary key id order


lib/rails_admin/support/csv_converter.rb
```diff
@@ -71,13 +71,22 @@ def export_field_for(method, model_config = @model_config)
...
     def generate_csv_string(options)
       generator_options = (options[:generator] || {}).symbolize_keys.delete_if { |_, value| value.blank? }
-      method = @objects.respond_to?(:find_each) ? :find_each : :each
 
       CSV.generate(**generator_options) do |csv|
         csv << generate_csv_header unless options[:skip_header]
 
-        @objects.send(method) do |object|
-          csv << generate_csv_row(object)
+        if @objects.respond_to?(:page)
```
kcromanpl, line 78:
```rb
if @objects.respond_to?(:page)
  page_num = 1
  loop do
    batch = @objects.page(page_num)
    break if batch.empty?

    batch.each { |object| csv << generate_csv_row(object) }
    page_num += 1
  end
else
  @objects.each { |object| csv << generate_csv_row(object) }
end
    batch = @objects.page(page_num)
    break if batch.empty?

    batch.each { |object| csv << generate_csv_row(object) }
    page_num += 1
  end
else
  @objects.each { |object| csv << generate_csv_row(object) }
end
```

- Replaced the while loop with a loop and break construct to make the code more readable and maintainable.
- Instead of using batch.any? to check if there are elements in the batch, Used batch.empty? This is more explicit and easier to understand.
- Moved the **page_num += 1** statement to the end of the loop to avoid repetition.

kcromanpl, line 78:
```rb
if @objects.respond_to?(:page)
  page_num = 1
  loop do
    batch = @objects.page(page_num)
    break if batch.empty?

    batch.each { |object| csv << generate_csv_row(object) }
    page_num += 1
  end
else
  @objects.each { |object| csv << generate_csv_row(object) }
end
```
-     Replaced the while loop with a loop and break construct to make the code more readable and maintainable.
-     Instead of using batch.any? to check if there are elements in the batch, Used batch.empty? This is more explicit and easier to understand.
-     Moved the page_num += 1 statement to the end of the loop to avoid repetition.

kcromanpl, line 78: Hi @pcai, minor refactor
```rb
if @objects.respond_to?(:page)
  page_num = 1
  loop do
    batch = @objects.page(page_num)
    break if batch.empty?

    batch.each { |object| csv << generate_csv_row(object) }
    page_num += 1
  end
else
  @objects.each { |object| csv << generate_csv_row(object) }
end
```
- Replaced the while loop with a loop and break construct to make the code more readable and maintainable.
- Instead of using batch.any? to check if there are elements in the batch, Used batch.empty? This is more explicit and easier to understand.
- Moved the page_num += 1 statement to the end of the loop to avoid repetition.

spec/rails_admin/support/csv_converter_spec.rb
```diff
@@ -151,5 +151,32 @@
...
+
+      let(:objects) { Player.all }
+      let(:options) { {} }
+
+      it 'generates a csv with all objects' do
+        expect(subject[2].split("\n").count).to be 101
+      end
+    end
+
+    context 'when objects are ordered' do
```
kcromanpl, line 168: If this is okay to you, please also consider matching whole records. This will make sure of order even for other code changes.
```rb
context 'when objects are ordered' do
  # or use before block as before
  let!(players) do
    FactoryBot.create_list :player, 2 do |player, index|
      player.name = "Player #{index}"
    end
  end

  let(:objects) { Player.all.order(name: :desc) }
  let(:options) { {} }

  # modify the subject block to return the array of objects
  it 'preserves the ordering' do
    expect(subject[2].split("\n")[1]).to eq('Player 1', 'Player 2')
  end
end
```

lib/rails_admin/support/csv_converter.rb
kcromanpl, line 78: Hi @pcai, minor refactor
```rb
if @objects.respond_to?(:page)
  page_num = 1
  loop do
    batch = @objects.page(page_num)
    break if batch.blank?

    batch.each { |object| csv << generate_csv_row(object) }
    page_num += 1
  end
else
  @objects.each { |object| csv << generate_csv_row(object) }
end
```
- Replaced the while loop with a loop and break construct to make the code more readable and maintainable.
- Instead of using batch.any? to check if there are elements in the batch, Used _batch.blank?_ or _batch.empty?_ This is more explicit and easier to understand.
- Moved the page_num += 1 statement to the end of the loop to avoid repetition.

spec/rails_admin/support/csv_converter_spec.rb
pcai: Thanks. I incorporated some of your suggestions, but matching the whole record is a bit nontrivial because the pk ids are not predictable when randomizing test seed

pcai:
I'm still looking for feedback on this. I don't mind keeping it up to date and passing CI, but I would appreciate some guidance on whether the change would be accepted at all.

mshibuya:
Thank you for this, and sorry it sat for so long.

#2580 is now fixed on master by #3752, which took a different route: `AbstractModel#each_record`, implemented per adapter, which `CsvConverter` iterates instead of materialising the relation.

Why this one wasn't merged as-is — all on the ActiveRecord side:

- `@objects.page(n)` uses the model's Kaminari `per_page`, so the batch size ends up being whatever the app happens to have configured, which can be very low.
- Offset paging with no tie-break: when the sort column has ties, nothing makes the order total, so rows can be skipped or repeated between pages.
- Deep `OFFSET` gets progressively more expensive as the export goes on.

What landed keeps batching but picks the strategy from the scope: `find_each` when the list has no explicit order, `find_each(order:)` when the list's order is the primary key order, offset paging with a primary-key tie-break otherwise, and plain `each` when the scope is already limited. Mongoid keeps its own iteration.

Closing as superseded — but the report and your push on it are what got this fixed, so thank you.

PR #3579, opened 2022-12-04
Remove unnecessary executable bit from Rakefile

jdufresne:


mshibuya:
Rubocop check is failing, how about removing the shebang?
https://github.com/railsadminteam/rails_admin/blob/997c1619cee3fb223b37e426d0ecdd54e941105e/spec/dummy_app/Rakefile#L1

jdufresne:
Thanks. Done in latest.

PR #3584, opened 2023-01-03
Remove support for end-of-life Ruby 2.6

jdufresne:
Ruby 2.6 has been EOL since 2022-04-12. It is no longer receiving security or bug fixes and its use is not supported. See: https://www.ruby-lang.org/en/downloads/branches/

Removing Ruby 2.6 will reduce the testing and maintenance required for this version. Workarounds can be removed. The RailsAdmin project can begin adopting newer Ruby coding conventions when convenient.

This removes jruby-9.3 from CI testing as that represents Ruby 2.6. jruby-9.4 was not added as that does not pass test suite.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,11 +14,6 @@ jobs:
         adapter: [sqlite3]
         asset: [webpack]
         include:
-          - ruby: 2.6
-            gemfile: gemfiles/rails_6.0.gemfile
-            orm: active_record
-            adapter: sqlite3
-            asset: sprockets
           - ruby: 2.7
             gemfile: gemfiles/rails_6.1.gemfile
             orm: active_record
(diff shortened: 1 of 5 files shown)


mshibuya:
Thanks for the PR, but again this is the policy I have been using for supported Ruby/Rails versions.
https://github.com/railsadminteam/rails_admin/pull/3430#issuecomment-982554274
 
And for this time, there appears to be no compelling reason to drop the 2.6 support. I agree that application developers should avoid using EOLed-Rubies, but this project is a library. We should be careful on dropping support for old versions, since if we go too strict users will suffer in a dependency hell (being stuck with contradicting requirements of their dependencies).

So my opinion is it's better to drop 2.6 support on the next major release (maybe on the release of Rails 8.0, unless no other major changes happen). What do you think about this?

jdufresne:
> So my opinion is it's better to drop 2.6 support on the next major release

Makes complete sense to me. Thanks.

PR #3585, opened 2023-01-03
Enable RuboCop Style/Encoding and apply auto-fixes

jdufresne:
Docs on the cop:
https://docs.rubocop.org/rubocop/cops_style.html#styleencoding

Docs on the Ruby style guide:
https://rubystyle.guide/#utf-8

> TIP: UTF-8 has been the default source file encoding since Ruby 2.0.


mshibuya:
👍

PR #3587, opened 2023-01-07
Restore caching in RailsAdmin::Config::Model#excluded?

jdufresne:
Partially reverts 2fa86c361b88fbd4c36adaad60be2104e57d75f5. The commit introduced a regression such that RailsAdmin::Config::Model#excluded? stopped caching false values.

### DIFF
--- a/lib/rails_admin/config/model.rb
+++ b/lib/rails_admin/config/model.rb
@@ -48,7 +48,9 @@ module RailsAdmin
       end
 
       def excluded?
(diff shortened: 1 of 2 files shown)


mshibuya:
Can this behavior tested in the spec?

jdufresne:
@mshibuya Good call. Amended an existing test in the latest push that demonstrates the caching.

mshibuya:
Thank you 👍

PR #3588, opened 2023-01-07
Switch to the safe navigation operator in spec files

jdufresne:
Ruby now has native support safe navigation so it can sometimes be used instead of the Rails Object#try method.

The change is limited to just the specs until other changes can be thoroughly verified. The safe navigation operator is different in that if the object is non-nil and the object does not respond to the method, a NoMethodError will be raised.


mshibuya:
👍

PR #3602, opened 2023-02-19
Fix image_html path

antontre1:
situation: Picture was not able to be correctly displayed in a record referring to an active storage picture

error identified: incorrect link given to show the picture in the view

fix provided:
replace `thumb_url` in `image_html = v.image_tag(thumb_url, class: 'img-thumbnail')` by `resource_url`


mshibuya:
What this does is just breaking the functionality of thumbnail feature. Please create an issue first, then we can talk about the solutions.

Issue #3565, opened 2022-11-17
LexerError: Lexical error on line 1: Unrecognized text.


andreierdoss:
**Describe the bug**
In production (Heroku) assets fail to compile due to a lack of interpolation for an SCSS variable. Locally it compiles without issues. From my research, it seems that the $avatar-size variable should be wrapped in #{} Similar issue was found [here](https://github.com/creativetimofficial/ct-light-bootstrap-dashboard-pro-react/issues/37).

**Reproduction steps**
git push heroku master

**Expected behavior**
Assets should compile with any issues. 

**Additional context**

- `rails` version: 6
- `rails_admin` version: 3.1.0
- `rails_admin` npm package version: 3.1.0
-  line code causing the [error](https://github.com/railsadminteam/rails_admin/blob/bb303772d74de6de8c99131de95cae070200248d/src/rails_admin/styles/base/theming.scss#L22) I noticed that in that file, some variables use interpolation others don't.
- full stack trace (if there's an exception)
```
remote:        JisonLexerError: Lexical error on line 1: Unrecognized text.
remote:
remote:          Erroneous area:
remote:        1: (40px - $avatar-size) / 2
remote:        ^..........^
remote:            at /tmp/build_0af411aa/css/rails_admin-86279344.css:14:202126
remote:            at Object.parseError (/tmp/build_0af411aa/node_modules/postcss-calc/dist/parser.js:1200:15)
remote:            at Object.lexer_parseError [as parseError] (/tmp/build_0af411aa/node_modules/postcss-calc/dist/parser.js:2333:44)
remote:            at Object.lexer_next [as next] (/tmp/build_0af411aa/node_modules/postcss-calc/dist/parser.js:3328:22)
remote:            at Object.lexer_fastLex [as fastLex] (/tmp/build_0af411aa/node_modules/postcss-calc/dist/parser.js:3403:18)
remote:            at fastLex (/tmp/build_0af411aa/node_modules/postcss-calc/dist/parser.js:1603:27)
remote:            at Parser.parse (/tmp/build_0af411aa/node_modules/postcss-calc/dist/parser.js:1677:30)
remote:            at /tmp/build_0af411aa/node_modules/postcss-calc/dist/lib/transform.js:33:30
remote:            at walk (/tmp/build_0af411aa/node_modules/postcss-calc/node_modules/postcss-value-parser/lib/walk.js:7:16)
[...]

PR #3607, opened 2023-03-08
Fix for heroku build error

greghaygood:
Heroku rejected my builds using rails_admin 3.1.1 until I added braces around these SCSS variables.

### DIFF
--- a/src/rails_admin/styles/base/theming.scss
+++ b/src/rails_admin/styles/base/theming.scss
@@ -19,10 +19,10 @@ body.rails_admin {
 
       img {
         position: absolute;
-        top: calc((40px - $avatar-size) / 2);
+        top: calc((40px - #{$avatar-size}) / 2);
 
         & + span {
-          margin-left: ($avatar-size + 5px);
+          margin-left: (#{$avatar-size} + 5px);
         }
       }
     }


mshibuya:
Could you try `yarn upgrade` first?
I suppose the issue you're having is the same one as #3565.

greghaygood:
Sorry for the delay - Yes, I tried yarn-upgrading with the official package in place, but no luck.

mshibuya:
Then please provide package.json and yarn.lock, then I can take a look.
This fix is very workaround-ish, so we should identify what's causing the issue before merging in.

greghaygood:
Sure, here you go. FYI I'm still getting the same error after updating to rails_admin 3.1.2.

Thanks for looking!

[package.json.txt](https://github.com/railsadminteam/rails_admin/files/11237270/package.json.txt)
[yarn.lock.txt](https://github.com/railsadminteam/rails_admin/files/11237271/yarn.lock.txt)

mshibuya:
Looks like your `@rails/webpacker` npm package is locked on the very old version. Could you try using the latest one 5.4.4, along with upgrading the `webpacker` gem version in Gemfile.
If the problem persists, please give me package.json, yarn.lock, and Gemfile again.

greghaygood:
The syntax error/warning I was getting went away, so upgrading the webpacker npm package seems to have fixed this issue. Heroku is now reporting a different compile warning during deployment, but it didn't block the build. So we can close this PR if you want. Thanks!

PR #3614, opened 2023-03-28
Use the same widget which works for nullable booleans also for non nullable

gabrieletassoni:
To preserve UI and UX consistency across all boolean-type fields, whether they are nullable or non-nullable.


mshibuya:
Thank you for the PR. Staged for the next major release.

mshibuya:
Merged in, finally!

PR #3616, opened 2023-04-06
Add support for '%-l' option to Flatpickr.

gguerini:
We currently support the options `%-d`, `%-H`, `%-I`, `%-M`, `%-m`, `%-S`, but not `%-l`. Adding `%-l` will prevent an error to be raised when the Ruby format uses this option.

Here's an example of how this could be used:

```ruby
> datetime = DateTime.new(2023, 4, 6, 9, 15)
=> Thu, 06 Apr 2023 09:15:00 +0000
> datetime.strftime("at %-l:%M %p")
=> "at 9:15 AM" # no extra pad before the hour
> datetime.strftime("at %l:%M %p")
=> "at  9:15 AM" # extra pad before the hour
```


mshibuya:
Thank you!

PR #3627, opened 2023-05-17
Vulnerability fix: validate return_to param using request.host

nicoayala:
## Context 
An internal team audits the Maxwell Rails apps occasionally, and in the last run, they discovered a vulnerability. Please take a look at the details bellow:

> An arbitrary redirect vulnerability occurs when an application URL redirects users to a 3rd-party, attacker-controlled website. As users recognize the application URL, they are likely to interact with the link. This can be used for phishing attacks against application users.
> 
> The rails_admin gem allows a redirect URL to be specified as a GET parameter. The validation of this URL does not properly check the redirect’s host, allowing it to be used as an arbitrary redirect.

## Solution
Validate the URL starts with the ~`request.host`~ `request.base_url`

### DIFF
--- a/app/controllers/rails_admin/main_controller.rb
+++ b/app/controllers/rails_admin/main_controller.rb
@@ -56,14 +56,11 @@ module RailsAdmin
     end
 
     def back_or_index
-      params[:return_to].presence && allowed_return_to?(params[:return_to]) && (params[:return_to] != request.fullpath) ? params[:return_to] : index_path
+      allowed_return_to?(params[:return_to].to_s) ? params[:return_to] : index_path
     end
 
     def allowed_return_to?(url)
-      return true if url.starts_with?(request.base_url)
-      return false unless url.to_s.start_with?('/')
-
-      !url.to_s.start_with?('//')
+      url != request.fullpath && url.start_with?(request.base_url, '/') && !url.start_with?('//')
     end
 
     def get_sort_hash(model_config)
--- a/app/controllers/rails_admin/main_controller.rb
+++ b/app/controllers/rails_admin/main_controller.rb
@@ -56,7 +56,14 @@ module RailsAdmin
     end
 
     def back_or_index
-      params[:return_to].presence && params[:return_to].starts_with?(request.base_url) && (params[:return_to] != request.fullpath) ? params[:return_to] : index_path
+      params[:return_to].presence && allowed_return_to?(params[:return_to]) && (params[:return_to] != request.fullpath) ? params[:return_to] : index_path
+    end
+
+    def allowed_return_to?(url)
+      return true if url.starts_with?(request.base_url)
+      return false unless url.to_s.start_with?('/')
+
+      !url.to_s.start_with?('//')
     end
 
     def get_sort_hash(model_config)
--- a/app/controllers/rails_admin/main_controller.rb
+++ b/app/controllers/rails_admin/main_controller.rb
@@ -56,8 +56,7 @@ module RailsAdmin
     end
 
     def back_or_index
-      params[:return_to].presence && params[:return_to].starts_with?(request.base_url) &&
-        (params[:return_to] != request.fullpath) ? params[:return_to] : index_path
+      params[:return_to].presence && params[:return_to].starts_with?(request.base_url) && (params[:return_to] != request.fullpath) ? params[:return_to] : index_path
     end
 
     def get_sort_hash(model_config)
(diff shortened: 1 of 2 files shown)


app/controllers/rails_admin/main_controller.rb
```diff
@@ -56,7 +56,7 @@ def respond_to_missing?(sym, include_private)
     end
 
     def back_or_index
-      params[:return_to].presence && params[:return_to].include?(request.host) && (params[:return_to] != request.fullpath) ? params[:return_to] : index_path
+      params[:return_to].presence && params[:return_to].starts_with?(request.host) && (params[:return_to] != request.fullpath) ? params[:return_to] : index_path
```
mshibuya, line 59: But `request.host` is in the form of `www.example.com`. So this won't allow the valid url `https://www.example.com/`, right?

.rubocop.yml
```diff
@@ -109,6 +109,8 @@ Metrics/BlockNesting:
 Metrics/ClassLength:
   CountComments: false
   Max: 132 # TODO: Lower to 100
+  Exclude:
+    - "lib/rails_admin/config.rb"
```
mshibuya, line 113: Thank you for taking care, but this issue got a different fix so please restore.
https://github.com/railsadminteam/rails_admin/commit/dec5be4fc48c1e626ef91796be0cb9fb61da1f7b

app/controllers/rails_admin/main_controller.rb
nicoayala: You are right, we need to take the scheme into account. I'll fix it. Thx 🙌 !

app/controllers/rails_admin/main_controller.rb
```diff
@@ -56,7 +56,14 @@ def respond_to_missing?(sym, include_private)
...
     def back_or_index
-      params[:return_to].presence && params[:return_to].include?(request.host) && (params[:return_to] != request.fullpath) ? params[:return_to] : index_path
+      params[:return_to].presence && allowed_return_to?(params[:return_to]) && (params[:return_to] != request.fullpath) ? params[:return_to] : index_path
+    end
+
+    def allowed_return_to?(url)
+      return true if url.starts_with?(request.base_url)
+      return false unless url.to_s.start_with?('/')
+
+      !url.to_s.start_with?('//')
```
nicoayala, line 66: Took inspiration from Rails7's [url_from](https://github.com/rails/rails/blob/cdd14ce1f5196e4bd98df42f89a1cd36ba9d4bee/actionpack/lib/action_controller/metal/redirecting.rb#L203-L212)

nicoayala (commented):
Ready for another round.

app/controllers/rails_admin/main_controller.rb
```diff
@@ -56,7 +56,14 @@ def respond_to_missing?(sym, include_private)
     end
 
     def back_or_index
-      params[:return_to].presence && params[:return_to].include?(request.host) && (params[:return_to] != request.fullpath) ? params[:return_to] : index_path
+      params[:return_to].presence && allowed_return_to?(params[:return_to]) && (params[:return_to] != request.fullpath) ? params[:return_to] : index_path
```
codealchemy, line 59: Minor - consolidating the logic for checking if the `return_to` is allowed in the extracted method would make this easier to follow.

```suggestion
      if _allowed_return_to?(params[:return_to].to_s) 
        params[:return_to]
      else
        index_path
      end
```

app/controllers/rails_admin/main_controller.rb
codealchemy: Tweaking this a bit to account for consolidation suggestion mentioned above, I believe this would cover the cases from the original and your changes.

```suggestion
    def _allowed_return_to?(url)
      url != request.fullpath &&
        (url.starts_with?(request.base_url, '/') && !url.start_with?('//'))
```

codealchemy (approved):

nicoayala:
The failing check is rubocop in:

```
Offenses:

lib/rails_admin/extensions/paper_trail/auditing_adapter.rb:149:28: C: [Correctable] Style/RedundantSelfAssignmentBranch: Remove the self-assignment branch.
          versions = all ? versions : versions.send(Kaminari.config.page_method_name, current_page).per(per_page)
                           ^^^^^^^^

444 files inspected, 1 offense detected, 1 offense autocorrectable
Error: Process completed with exit code 1.
```

Is this being addressed? Let me know how should I proceed

mshibuya:
Merged in. Thank you for the fix!

mshibuya:
The Rubocop failure was already addressed in the master:
f06df8ff7f9890d7cf60fd4b724a4d92f55e94ab

PR #3630, opened 2023-06-12
Reset polymorphic id upon type change

KMinChrist:
Hi, thank you for your attention. Currently, for polymorphic association fields, the id persists after type is changed, which may be unclear to users and cause a bug if the id is selected before type. With this change, the id field can be reset each time the type field changes. I hope you can approve this merge request. Thank you!

### DIFF
--- a/src/rails_admin/widgets.js
+++ b/src/rails_admin/widgets.js
@@ -258,6 +258,7 @@ import I18n from "./i18n";
             "options"
           );
           object_select.data("options", selected_data);
+          object_select.val('');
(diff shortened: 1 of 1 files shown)


mshibuya:
Thank you for the PR. The fix makes sense, could you add a test case to verify this?
https://github.com/railsadminteam/rails_admin/blob/master/spec/integration/fields/polymorphic_assosiation_spec.rb will be the best place to work.

mshibuya:
Let me write a test. Thank you.

PR #3636, opened 2023-07-06
Replace &rdquo; entity with inline <q> element

jdufresne:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/q

> Most modern browsers implement this by surrounding the text in
> quotation marks.

Importantly, this uses the localized quotation glyph. In English, this is the same left/right quotation mark.


codealchemy (approved):

codealchemy:
Thanks!

PR #3643, opened 2023-08-14
Integration with ViteRuby

bankair:
Hey folks,

Here, at Simundia, we are using rails_admin, and we are loving it so far. Congrat's on the work you did on this.

Since we are currently in the process of migrating our assets pipeline from webpack to vite, we made a small fork which adapt the generators and the engine in order to be able to work with [ViteRuby](https://vite-ruby.netlify.app/) as well.

However, we don't want to maintain this fork forever, and we are wondering if you'd be keen to merge our diff with rails_admin's trunk.

Please tell us if you need anything from us in order to move forward with it.

Have a great day!


mshibuya:
Great work! Some more changes are needed, but I can take care of them.
Thank you so much 🏅

bankair:
Great, thank you for taking care of the extra-mile!

PR #3649, opened 2023-10-06
Stop using update_only to decide to show subform on create

### LINKED ISSUE #2626: fields.nested_form[:update_only] should always be false by default.
See the "first bug" in [this stack overflow answer](http://stackoverflow.com/a/36929302/1877609). 

It's a personal opinion, but I think that interpreting `accept_nested_attributes_for: ..., update_only:true` as "disable the form when creating the object" is a potential misunderstanding of the option, (for Rails 4), and should at least be documented in the wiki?

myers:
Closes #2626 

There seems to be some confusion here about what `:update_only` means in `accepts_nested_attributes_for` in Rails Admin.  It is currently use to decide to show the subform for a child record when creating a new record.  I don't think this is correct. 

https://api.rubyonrails.org/classes/ActiveRecord/NestedAttributes/ClassMethods.html

> For a one-to-one association, this option allows you to specify how nested attributes are going to be used when an associated record already exists. In general, an existing record may either be updated with the new set of attribute values or be replaced by a wholly new record containing those values. By default the :update_only option is false and the nested attributes are used to update the existing record only if they include the record’s :id value. Otherwise a new record will be instantiated and used to replace the existing one. However if the :update_only option is true, the nested attributes are used to update the record’s attributes always, regardless of whether the :id is present. The option is ignored for collection associations.

This PR no longer attempts to use `update_only`, and only looks at `field.inline_add`.

Also as a bonus removing an `unless` with `||` makes my [Grug brain](https://grugbrain.dev/) happy.

### DIFF
--- a/app/views/rails_admin/main/_form_nested_many.html.erb
+++ b/app/views/rails_admin/main/_form_nested_many.html.erb
@@ -3,7 +3,7 @@
     <a class="<%= (field.active? ? 'active' : '') %> btn btn-info toggler" data-bs-target="<%= form.jquery_namespace(field) %> .collapse" data-bs-toggle="collapse" role="button">
       <i class="fas"></i>
     </a>
-    <% unless field.nested_form[:update_only] || !field.inline_add %>
+    <% if field.inline_add %>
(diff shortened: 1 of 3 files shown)


mshibuya:
Totally makes sense, thank you for the PR 👍

PR #3652, opened 2023-10-13
Quote table_ and column_name for sorting

### LINKED ISSUE #1631: Table names don't always get properly quoted
In lib/rails_admin/config/actions/dashboard.rb, on line 30, it produces an SQL query that doesn't quote the table name referenced (SELECT  "user".\* FROM "user"  ORDER BY user.updated_at desc). This is a problem if you have a table name like "user", which is a reserved word in PostgreSQL.

Here's a gist with a stack trace: https://gist.github.com/jasonswett/1da12e239266f20b17c2

I wrote a fix for this, but this would be my first open source contribution ever, so I might need a little hand-holding to get it in. I understand I'm supposed to write a failing spec, but I don't see a spec for dashboard.rb in what I would assume to be the logical place, spec/rails_admin/config/actions. If someone could help me find that, that would be appreciated.

rnestler:
Fixes #1631.

### DIFF
--- a/lib/rails_admin/abstract_model.rb
+++ b/lib/rails_admin/abstract_model.rb
@@ -61,11 +61,19 @@ module RailsAdmin
     end
 
     def quoted_table_name
-      model.quoted_table_name
+      if adapter == :active_record
+        model.quoted_table_name
+      else
+        table_name
+      end
     end
 
     def quote_column_name(name)
-      model.connection.quote_column_name(name)
+      if adapter == :active_record
+        model.connection.quote_column_name(name)
+      else
+        name
+      end
     end
 
     def to_s
--- a/lib/rails_admin/abstract_model.rb
+++ b/lib/rails_admin/abstract_model.rb
@@ -60,6 +60,14 @@ module RailsAdmin
       @model_name.constantize
     end
 
+    def quoted_table_name
+      model.quoted_table_name
+    end
+
+    def quote_column_name(name)
+      model.connection.quote_column_name(name)
+    end
+
     def to_s
       model.to_s
     end
--- a/lib/rails_admin/abstract_model.rb
+++ b/lib/rails_admin/abstract_model.rb
@@ -61,19 +61,11 @@ module RailsAdmin
     end
 
     def quoted_table_name
-      if adapter == :active_record
-        model.quoted_table_name
-      else
-        table_name
-      end
+      table_name
     end
 
     def quote_column_name(name)
-      if adapter == :active_record
-        model.connection.quote_column_name(name)
-      else
-        name
-      end
+      name
     end
 
     def to_s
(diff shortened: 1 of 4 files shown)


spec/controllers/rails_admin/main_controller_spec.rb
```diff
@@ -73,14 +73,14 @@ def get(action, params)
...
-        expect(controller.send(:get_sort_hash, RailsAdmin.config(Player))).to eq(sort: 'players.team_id', sort_reverse: true)
+        expect(controller.send(:get_sort_hash, RailsAdmin.config(Player))).to match(sort: /.?players.?\..?team_id.?/, sort_reverse: true)
       end
     end
 
     context 'using active_record, supporting joins', active_record: true do
       it 'gives back the local column' do
         controller.params = {sort: 'team', model_name: 'players'}
-        expect(controller.send(:get_sort_hash, RailsAdmin.config(Player))).to eq(sort: 'teams.name', sort_reverse: true)
+        expect(controller.send(:get_sort_hash, RailsAdmin.config(Player))).to match(sort: /.?teams.?\..?name.?/, sort_reverse: true)
```
rnestler, line 83: Maybe I should also add a regression spec with a model with capitalized fields? Where would I put it?

rnestler:
Any update on this?

lib/rails_admin/abstract_model.rb
```diff
@@ -60,6 +60,22 @@ def model
       @model_name.constantize
     end
 
+    def quoted_table_name
+      if adapter == :active_record
```
mshibuya, line 64: Instead of putting a conditional here, you can create the default implementation

```
    def quoted_table_name
      table_name
    end
```

and move the ActiveRecord specific part into [`RailsAdmin::Adapters::ActiveRecord`](https://github.com/railsadminteam/rails_admin/blob/master/lib/rails_admin/adapters/active_record.rb).

spec/controllers/rails_admin/main_controller_spec.rb
mshibuya: Somewhere around here will be good, as this fix is about `#get_sort_hash`.
Also you can create RDBMS-specific test cases like:
https://github.com/railsadminteam/rails_admin/blob/0e12e5b465997c14d5b7a4d500a0d4cebed21aa9/spec/rails_admin/support/csv_converter_spec.rb#L69-L75

mshibuya (commented):
Thanks for the PR! I've made some comments.

lib/rails_admin/abstract_model.rb
rnestler: Fixed.

mshibuya:
Great job, thank you!

rnestler:
Ah sorry I forgot about that. I think I didn't add the regression test or RDBMS-specific test cases.

mshibuya:
I'll handle that when I have time 👍

PR #3656, opened 2023-11-15
Show a thumbnail of all representable Active Storage files + images

wvengen:
Currently, only images have a preview when using Active Storage attachments. Preview of videos (and potentially other file formats) are supported, however, but one needs to use _representable_ instead of only _previewable_. This PR switches to using _representable_.

Tested with movie previews and images on multiple upload.

Note that we still need to look at the image mime-type, to show e.g. SVG images (which are not representable, but can still be shown directly - great that there is an automated test for this).

See also:
- https://edgeguides.rubyonrails.org/active_storage_overview.html
- https://edgeapi.rubyonrails.org/classes/ActiveStorage/Blob/Representable.html

### DIFF
--- a/lib/rails_admin/config/fields/types/active_storage.rb
+++ b/lib/rails_admin/config/fields/types/active_storage.rb
@@ -18,10 +18,7 @@ module RailsAdmin
           end
 
           register_instance_option :image? do
-            if value
-              mime_type = Mime::Type.lookup_by_extension(value.filename.extension_without_delimiter)
-              mime_type.to_s.match?(/^image/)
-            end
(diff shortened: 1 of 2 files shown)


wvengen:
It's not clear to me why one test fails (jruby-9.3, rails_6.1, active_record, mysql2, sprockets) and all the others succeed. It doesn't look like it would be related to the changes in this PR.

mshibuya:
Tests are required, but I can handle that. Thank you.

wvengen:
Thank you, also for the last bits!

Issue #3661, opened 2023-12-08
Support for turbo-rails 2.x gem


ArturT:
Turbo 8 has been released in the turbo-rails gem [`2.0.0.pre.beta.1`](https://rubygems.org/gems/turbo-rails/versions/2.0.0.pre.beta.1).

Turbo 8 has a few [cool features](https://fly.io/ruby-dispatch/turbo-8-in-8-minutes/) and I wanted to test this with the Rails app but the rails_admin depends on turbo-rails < 2 (see `spec.add_dependency 'turbo-rails', '~> 1.0'` in `rails_admin.gemspec`).

**Reproduction steps**

I [updated the rails_admin.gemspec](https://github.com/KnapsackPro/rails_admin/commit/c96a6642b3ce17eab5f62c7297c1fd12c478226e) file to:

```ruby
# rails_admin.gemspec
spec.add_dependency 'turbo-rails', ['>= 1.0', '< 3']
```

I was able to run the Rails app with rails_admin and the latest turbo-rails.
I'm using turbo-rails directly from github because they did some fix ([the bug prevented to start my Rails app](https://github.com/hotwired/turbo-rails/issues/527) fix: https://github.com/hotwired/turbo-rails/pull/523 ) that has not been released to rubygems yet.

```ruby
# my rails app
# Gemfile
gem 'turbo-rails', github: 'hotwired/turbo-rails', ref: '4eb4e928e30be8cd537af8073f98b80ddea4a578'
```

**Describe the bug**

When it comes to rails_admin. While I'm using Turbo 8 (turbo-rails 2.x gem) I found an issue.
When I visit the http://localhost:3000/admin/user
In the search bar I type a user name and click on Refresh button then nothing happens. I get the error in the Chrome console:

```
RangeError: Maximum call stack size exceeded
```

![Screenshot 2023-12-08 at 17 25 47](https://github.com/railsadminteam/rails_admin/assets/662545/985d5914-2013-4026-9623-264ca83097b6)
![Screenshot 2023-12-08 at 17 25 59](https://github.com/railsadminteam/rails_admin/assets/662545/692a3a90-6a0a-443a-8e47-07f693d7a6fb)

**Workaround**

In order to keep using rails_admin and the latest turbo-rails 2.x I disabled Turbo in the rails_admin dashboard.

I added:

```js
// app/assets/javascripts/rails_admin/custom/ui.js
window.Turbo.session.drive = false
```

[...]

PR #3671, opened 2024-02-05
Allow turbo-rails 2 in gemspec

bb:
See https://github.com/railsadminteam/rails_admin/issues/3661#issuecomment-1927663367

### DIFF
--- a/rails_admin.gemspec
+++ b/rails_admin.gemspec
@@ -11,7 +11,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'kaminari', '>= 0.14', '< 2.0'
   spec.add_dependency 'nested_form', '~> 0.3'
   spec.add_dependency 'rails', ['>= 6.0', '< 8']
-  spec.add_dependency 'turbo-rails', '~> 1.0'
+  spec.add_dependency 'turbo-rails', ['>= 1.0', '< 3']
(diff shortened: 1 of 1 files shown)


pathakh (approved):

mshibuya:
CI is failing, but I can handle it. Thanks!

bb:
Thank you!

xijo:
@pathakh any chance to release this soon-ish? Feels bad to force us all to stay on an outdated version of turbo :|

bb:
@mshibuya thanks for providing the hotfixes 3.1.3+4. Are there any plans to create a new minor release from master (which includes this PR)?

If not, could you please update the patch version in master to 3.1.4 so we can refer to master in our bundle without having to ignore https://github.com/railsadminteam/rails_admin/security/advisories/GHSA-8qgm-g2vv-vwvc in our audits.

mshibuya:
I've just released a pre-release version 3.2.0.beta. The master was also bumped to that version.
It'll take some more time for the next minor release, but please give feedback so it can happen earlier 🙏

bb:
That's great, thank you! I tried it locally and in CI. Also converted an App from Sprockets to Ruby-Vite (thanks to #3643) which also works for me. Good to go, from my perspective. 🚀

PR #3672, opened 2024-02-07
Correct meta tags syntax

coorasse:
meta tags are void elements, so this is the correct syntax for them. See also: https://developer.mozilla.org/en-US/docs/Glossary/Void_element

### DIFF
--- a/app/views/layouts/rails_admin/_head.html.erb
+++ b/app/views/layouts/rails_admin/_head.html.erb
@@ -1,7 +1,7 @@
-<meta content="IE=edge" http-equiv="X-UA-Compatible"/>
-<meta content="text/html; charset=utf-8" http-equiv="Content-Type"/>
-<meta content="width=device-width, initial-scale=1" name="viewport; charset=utf-8"/>
-<meta content="NONE,NOARCHIVE" name="robots"/>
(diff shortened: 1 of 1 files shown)


codealchemy (approved):
👍

app/views/layouts/rails_admin/_head.html.erb
```diff
@@ -1,7 +1,7 @@
-<meta content="IE=edge" http-equiv="X-UA-Compatible"/>
-<meta content="text/html; charset=utf-8" http-equiv="Content-Type"/>
-<meta content="width=device-width, initial-scale=1" name="viewport; charset=utf-8"/>
-<meta content="NONE,NOARCHIVE" name="robots"/>
+<meta content="IE=edge" http-equiv="X-UA-Compatible">
+<meta content="text/html; charset=utf-8" http-equiv="Content-Type">
+<meta content="width=device-width, initial-scale=1" name="viewport; charset=utf-8">
+<meta content="NONE,NOARCHIVE" name="robots">
```
rnestler, line 4: In the examples at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#examples it uses the self-closing tags though:

image

mshibuya:
Thanks!

PR #3676, opened 2024-03-01
Fix default config path for  ImportmapFormatter

rnestler:


rnestler:
I think the CI failure is not related to this change, since it also happens on current master.

mshibuya:
Thank you 😃

Issue #3651, opened 2023-10-09
Error When Filtering Legacy Enum Fields


deepakisrani:
**Description**
When applying the filters on the fields populated using the legacy enum, rails_admin breaks throwing the `undefined method [] for nil:NilClass` while rendering the content on the page.

We use rails_admin in a large Rails project and started facing this issue after upgrading rails_admin from 2 to 3 and the Rails version from 6 to 7. I have created a dummy application replicating the setup we have with all the configuration, first we had assumed that the problem lies within the configuration of our company's code but the issue replicated itself within the new application as well, [link](https://github.com/deepakisrani/test_app) to the repo, this was a fresh install of Rails 7.1 along with rails_admin.

**Reproduction Steps**
Apply a filter on any of the fields utilizing the legacy enum.

**Expected Behavior**
Expect the page to render properly with either the resulting set or no records found.

**Additional context**

- `rails` version: 7.1
- `ruby` version: 2.7.6
- `rails_admin` version: 3.1.2
- `rails_admin` npm package version: NA (Using sprockets)
- Github Link to dummy application: https://github.com/deepakisrani/test_app
- full stack trace

```
[rails_admin (3.1.2) lib/rails_admin/config/fields/types/enum.rb:31:in `block in <class:Enum>'](http://localhost:3000/admin/enum_test?f%5Bsingle_value%5D%5B62405%5D%5Bv%5D=Key+2&query=#)
[rails_admin (3.1.2) lib/rails_admin/config/configurable.rb:77:in `instance_eval'](http://localhost:3000/admin/enum_test?f%5Bsingle_value%5D%5B62405%5D%5Bv%5D=Key+2&query=#)
[rails_admin (3.1.2) lib/rails_admin/config/configurable.rb:77:in `block in register_instance_option'](http://localhost:3000/admin/enum_test?f%5Bsingle_value%5D%5B62405%5D%5Bv%5D=Key+2&query=#)
[rails_admin (3.1.2) lib/rails_admin/config/fields/types/enum.rb:14:in `block in <class:Enum>'](http://localhost:3000/admin/enum_test?f%5Bsingle_value%5D%5B62405%5D%5Bv%5D=Key+2&query=#)
[rails_admin (3.1.2) lib/rails_admin/config/configurable.rb:77:in `instance_eval'](http://localhost:3000/admin/enum_test?f%5Bsingle_value%5D%5B62405%5D%5Bv%5D=Key+2&query=#)
[...]

PR #3647, opened 2023-09-27
Fix filter on enum value in list view

fuegas:
Filtering on enum (in list view) gave the following error:
```
ActionView::Template::Error (undefined method `[]' for nil:NilClass

              (bindings[:object] || abstract_model.model.new).send(enum_method)
                       ^^^^^^^^^):
    54: <div id="list">
    55:   <%= form_tag(index_path(params.except(*%w[page f query])), method: :get) do %>
    56:     <div class="card mb-3 p-3 bg-light">
    57:       <div class="row" data-options="<%= ordered_filter_options.to_json %>" id="filters_box"></div>
    58:       <hr class="filters_box" style="display:<%= ordered_filters.empty? ? 'none' : 'block' %>"/>
    59:       <div class="row">
    60:         <div class="col-sm-6">

rails_admin (3.1.2) lib/rails_admin/config/fields/types/enum.rb:31:in `block in <class:Enum>'
rails_admin (3.1.2) lib/rails_admin/config/configurable.rb:77:in `instance_eval'
rails_admin (3.1.2) lib/rails_admin/config/configurable.rb:77:in `block in register_instance_option'
rails_admin (3.1.2) lib/rails_admin/config/fields/types/enum.rb:14:in `block in <class:Enum>'
rails_admin (3.1.2) lib/rails_admin/config/configurable.rb:77:in `instance_eval'
rails_admin (3.1.2) lib/rails_admin/config/configurable.rb:77:in `block in register_instance_option'
rails_admin (3.1.2) lib/rails_admin/config/fields/base.rb:106:in `filter_options'
rails_admin (3.1.2) app/helpers/rails_admin/main_helper.rb:52:in `block in ordered_filter_options'
rails_admin (3.1.2) app/helpers/rails_admin/main_helper.rb:44:in `map'
rails_admin (3.1.2) app/helpers/rails_admin/main_helper.rb:44:in `ordered_filter_options'
rails_admin (3.1.2) app/views/rails_admin/main/index.html.erb:57
actionview (7.0.8) lib/action_view/helpers/capture_helper.rb:45:in `block in capture'
actionview (7.0.8) lib/action_view/helpers/capture_helper.rb:209:in `with_output_buffer'
actionview (7.0.8) lib/action_view/helpers/capture_helper.rb:45:in `capture'
actionview (7.0.8) lib/action_view/helpers/form_tag_helper.rb:78:in `form_tag'
rails_admin (3.1.2) app/views/rails_admin/main/index.html.erb:55
```

[...]

PR #3684, opened 2024-06-07
Fix filter on enum value in list view v2

robotfelix:
Filtering on legacy enums in a list view gives errors under Rails Admin 3.x (see #3651)

This PR adds to #3647 by adding a test for the fix provided there, as well as providing a similar fix for the `Fields::Types::Enum#enum_method` method.

I've left the original commit by @fuegas as-is to give due credit, but happy to squash the fix with the associated tests if preferred.

I don't fully grasp _why_ the method is called without bindings (not even an empty hash), and it's not clear if that is something that should ever be expected or is a symptom of a bug somewhere else, however either way it is true that the code paths using `abstract_model` were not covered by the existing tests.

### DIFF
--- a/lib/rails_admin/config/fields/types/enum.rb
+++ b/lib/rails_admin/config/fields/types/enum.rb
@@ -21,7 +21,7 @@ module RailsAdmin
           end
 
           register_instance_option :enum_method do
-            @enum_method ||= bindings[:object].class.respond_to?("#{name}_enum") || (bindings[:object] || abstract_model.model.new).respond_to?("#{name}_enum") ? "#{name}_enum" : name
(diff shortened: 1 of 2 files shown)


mshibuya:
For now I'm not merging this as https://github.com/railsadminteam/rails_admin/commit/d62f604cc8d7d1434f7dfe0c5aca3aaf3dc2547b will be the proper fix, but I really appreciate your attempt to address this.
Thank you so much, and looking forward to another contribution 👍

robotfelix:
@mshibuya Thanks. https://github.com/railsadminteam/rails_admin/commit/d62f604cc8d7d1434f7dfe0c5aca3aaf3dc2547b definitely looks like a better fix! 🎉 

Let me know if you'd like me to adapt the tests into some "when the binding don't provide an object" test cases (eg: with an empty binding rather than no binding) instead to add test coverage for the code paths through `abstract_model`.

PR #3702, opened 2024-11-08
Test against rails 8.0

jewilmeer:
As rails 8 came out, and not a lot has changed internally since 7.2, it might just work :-)


gemfiles/rails_8.0.gemfile
```diff
@@ -0,0 +1,55 @@
...
+  gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+end
+
+group :active_record do
+  gem "paper_trail", ">= 12.0"
+
+  platforms :ruby, :mswin, :mingw, :x64_mingw do
+    gem "mysql2", ">= 0.3.14"
+    gem "pg", ">= 1.0.0"
+    gem "sqlite3", "~> 1.3"
```
k-tsuchiya-jp, line 51: > https://github.com/railsadminteam/rails_admin/actions/runs/11738544873/job/32701267375?pr=3702#step:7:30

From Rails 8.0, it looks like the version of sqlite3 needs to be upgraded to v2.1 or higher.

jewilmeer: indeed, I upgraded it, and now it seems to work

jewilmeer:
It looks like a lot of unrelated things are broken, not sure how to proceed.

dsinn:
@jewilmeer Indeed, I was able to reproduce the issue on master. I suspect that because the [json gem](https://github.com/ruby/json) isn't pinned, the latest CI runs install a version that is greater than 2.7.6, the newer versions' `JSON.pretty_generate({})` behaves differently, as it no longer contains a newline. Compare:

<table>
<tr><th>Old behaviour</th><th>New behaviour</th></tr>
</tr>
<tr>
<td><pre>
$ irb
irb(main):001> require "json"
=> true
irb(main):002> JSON::VERSION
=> "2.7.6"
irb(main):003> puts JSON.pretty_generate({})
{
}
=> nil
</pre></td>
<td><pre>
$ irb
irb(main):001> require "json"
=> true
irb(main):002> JSON::VERSION
=> "2.8.2"
irb(main):003> puts JSON.pretty_generate({})
{}
=> nil

</pre></td>
</tr></table>

The test fails because RailsAdmin's Json field class [calls JSON.pretty_generate](https://github.com/railsadminteam/rails_admin/blob/b09f770d9f2b743626d5f29ac9449dbcce9e5c70/lib/rails_admin/config/fields/types/json.rb#L4#L15), and the test's regex is expecting at least one newline from its result. As a quick test, I made the following change, and the tests pass:

```diff
diff --git a/spec/rails_admin/config/fields/types/json_spec.rb b/spec/rails_admin/config/fields/types/json_spec.rb
index a1e03958..a749d930 100644
--- a/spec/rails_admin/config/fields/types/json_spec.rb
+++ b/spec/rails_admin/config/fields/types/json_spec.rb
@@ -24,7 +24,7 @@ RSpec.describe RailsAdmin::Config::Fields::Types::Json do
     it 'returns correct value for empty json' do
       allow(object).to receive(:json_field) { {} }
       actual = field.with(bindings).formatted_value
-      expect(actual).to match(/{\n+}/)
+      expect(actual).to match(/{\n*}/)
     end
 
     it 'retuns correct value' do
@@ -72,7 +72,7 @@ RSpec.describe RailsAdmin::Config::Fields::Types::Json do
     it 'returns correct value for empty json' do
       allow(object).to receive(:json_field) { {} }
       actual = field.with(bindings).export_value
-      expect(actual).to match(/{\n+}/)
+      expect(actual).to match(/{\n*}/)
     end
 
     it 'returns correct value' do
```

I don't see any need for the newline, so I'd lean towards just updating the test (whether to split the change into a separate PR is up to you and the maintainers), but I'm also not a frequent user of RailsAdmin.

jewilmeer:
@dsinn Thanks for figuring this out. I'm not sure what the idea of the gem versioning of the gem is. Pinning certain gems at a higher version might force some of the gem users to upgrade. Not sure if the idea is to keep it as wide as possible.

But I'm also not sure if the json spec makes a lot of sense. We might want to test if the result is parseable as JSON, instead of matching against a string with regexp.

Gemfile
```diff
@@ -49,7 +49,7 @@ group :active_record do
   platforms :ruby, :mswin, :mingw, :x64_mingw do
     gem 'mysql2', '>= 0.3.14'
     gem 'pg', '>= 1.0.0'
-    gem 'sqlite3', '~> 1.3'
+    gem 'sqlite3', '~> 2.2'
```
mshibuya, line 52: Changing a Gemfile entry causes every files under `gemfiles/` to change, once you run `appraisal generate`. If your intention is to use sqlite3 ~> 2.2 only for Rails 8, you need to move this to Appraisal so other Rails versions can stay on sqlite 1.x.

mshibuya:
I support the JSON spec change https://github.com/railsadminteam/rails_admin/pull/3702#issuecomment-2477676433 proposed by @dsinn, as the existence of newline is irrelevant to the purpose of the spec here.

Gemfile
jewilmeer: I've upgraded the sqlite version for all.

mshibuya: You shouldn't. That's why builds for older Rails versions are failing.
https://github.com/railsadminteam/rails_admin/actions/runs/11931789920/job/33255380343?pr=3702

gemfiles/rails_8.0.gemfile
k-tsuchiya-jp: @ jewilmeer
Thank you for your support with Rails 8.0! It’s really helpful!

> Seems like sqlite3 2.2 requires ruby 3.1.

This might help you.
In RailsAdmin, we use `gem 'appraisal'` to manage gems for each Rails version.
https://github.com/thoughtbot/appraisal?tab=readme-ov-file#setup

For versions other than Rails 8.0, there are no issues with the traditional gem versions, and no changes to Gemfile are necessary.
Instead, since https://github.com/railsadminteam/rails_admin/commit/bee353940fd79b85e914cd6b85ff024c9caf8182#diff-4fc39a0aadaf0ef5048e145a93b08c08740e7efd36c6890426664cea221a622eR51 is explicitly mentioned, this alone should make the Rails 8.0 tests work.

> What would be the best approach? Require ruby 3.1 for the next release, or stay backwards compatible?

This is just my opinion, but since Ruby 3.1 hasn't reached EOL yet, I think it's fine to include Ruby 3.1 in the next release.

Gemfile
k-tsuchiya-jp: @jewilmeer 
Please check, as this https://github.com/railsadminteam/rails_admin/pull/3702#discussion_r1851135751 is linked with @mshibuya's comment.

gemfiles/rails_8.0.gemfile
jewilmeer: Makes sense. I've changed the Gemfile back to the older sqlite3 version, and in the appraisal file I've made an exception for rails 8

jewilmeer:
I've managed to update to sqlite3 2.2 only for rails 8, but the dummy app in the specs is not happy with in. I don't have the knowledge to resolve it myself. Can any of you help out?

mshibuya:
OK I can take care of the rest. Thanks.

k-tsuchiya-jp:
@mshibuya 
Thank you for your contribution.
Could you kindly let me know when it will be released?

mshibuya:
Just released 3.3.0, which includes this change 🚀

PR #3703, opened 2024-11-19
Dropdown not being updated to dynamic options through custom js, overridden to default.

upendra-azad:
Not able to overwrite dropdown from version 3.2 onwards, because of default here 
https://github.com/railsadminteam/rails_admin/blob/b09f770d9f2b743626d5f29ac9449dbcce9e5c70/src/rails_admin/filtering-select.js#L300

it's done in (https://github.com/railsadminteam/rails_admin/pull/3630).

my use case is that i am overwriting the states picklist based on country picklist through custom js.
and to make the new dropdown work after updating the options, i am using `$(selector).filteringSelect('destroy')` to reinitialize. but its being overwritten with that line, we can add one boolean flag to ignore that behaviour.


upendra-azad:
Hi, Thanks for you attention, let me know if this is worth merging or not, so i can close this pr.

mshibuya:
Thanks for the PR. Could you please show me your custom js?
I want to understand your usecase correctly. There can be a better way to achieve what you want, maybe.

upendra-azad:
> Thanks for the PR. Could you please show me your custom js? I want to understand your usecase correctly. There can be a better way to achieve what you want, maybe.

Yeah sure, here is the sample code from erb file which i am using for state selection based on country selected dynamically.


```
<!--if "K" is the record class name, (k.country) k_country is already there, this is for (k.state) k_state -->

<%= form.select(:state, value_from_db(k), {include_blank: true}) %>

<script type="text/javascript">
  let allStates = <%= {"usa"=>[["alaska", "ak"]], "rome"=>[]}.to_json(root: false).html_safe %>;

  $(function() {
    let countrySelector = document.querySelector("#k_country");
    countrySelector.addEventListener("change", function () {

      const selectedCountry = $("#k_country").val();
      const stateSelector = $("#k_state");

      stateSelector.empty(); // Clear existing options

      if (selectedCountry) {
        const statesForCountry = allStates[selectedCountry];

        if (!statesForCountry || statesForCountry.length === 0) {
          // Add a blank option if no states are available
          stateSelector.append($("<option>").val("").text(""));
        } else {
          // Populate states for the selected country
          statesForCountry.forEach(([sn, sv]) => {
            stateSelector.append($("<option>").val(sn).text(sv));
          });
        }

        refreshPicklist("#k_state"); // Refresh the picklist
      }
      
    });
  });


  function refreshPicklist(selector) {
    if ($(selector).siblings('div').length > 0) {
      // it does not refresh (picklist does not show) without this
      try { $(selector).filteringSelect('destroy'); } catch (e) { }
      $(selector).filteringSelect();
    }
  }
</script>

```

mshibuya:
The problem is purely in your usage of the filtering-select widget.

```js
      try { $(selector).filteringSelect('destroy'); } catch (e) { }
      $(selector).filteringSelect();
```

What this does is not refreshing, but it destroys the filtering-select widget and then re-creates.
So it is totally legitimate that the options update you made gets lost on destroy. You should change your javascript logic to destroy the widget first, then apply changes to the options, finally re-create the widget.

Also if your lists of countries and states are stored in DB, client-side dynamic scoping feature that was added recently will fit well with this sort of use-cases.
https://github.com/railsadminteam/rails_admin/commit/12715f2dd12d97f0676e548e4271906df424b89d

upendra-azad:
@mshibuya i will try this, thanks a lot for your help.

PR #3705, opened 2024-12-19
Change default response format of the show action to HTML

jacob-carlborg-apoex:
This is consistent with the other actions. When making HTTP requests using tools that are not web browsers, like `curl` or `fetch` in JavaScript without an explicit `Accept` header Rails will receive the request as with the `Accept` header set to `*/*`. That means that all registered mime types are a potential match and Rails will pick the response type based on the order they're declared in the source code.


mshibuya:
Thanks!

PR #3717, opened 2025-03-19
Add belongs_to optional/required support to required?

RaymondFFT:
Active Record 7.1.0.beta1 added a lambda to the presence validation added by a belongs_to association. This lambda is only an optimisation to not check presence if the column is unchanged, however it stops RailsAdmin from treating the column as required.


mshibuya:
Sorry for taking time, thank you!

PR #3719, opened 2025-06-07
Add nonce to css/js tags

oz-tal:
Ensure css/js tags comply with the CSP nonce (if one is defined).

In the absence of a nonce (or the whole policy itself even), theses changes will gracefully do nothing.


mshibuya:
Thanks!

PR #3734, opened 2026-03-15
Upgrade @hotwired/turbo-rails to resolve security warning

jdufresne:
Details of the security vulnerability are available at:

https://github.com/hotwired/turbo/security/advisories/GHSA-qppm-g56g-fpvp

> A race condition in Turbo Frames allows delayed HTTP responses to
> restore stale session cookies after session-modifying operations.

As a result, my project started receiving the following warning:

> Dependabot cannot update @hotwired/turbo to a non-vulnerable version
>
> The latest possible version that can be installed is 7.3.0 because of the following conflicting dependencies:
>
> rails_admin@3.3.0 requires @hotwired/turbo@^7.3.0 via @hotwired/turbo-rails@7.3.0
> No patched version available for @hotwired/turbo
>
> The earliest fixed version is 8.0.21.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   "dependencies": {
     "@babel/runtime": "^7.16.7",
     "@fortawesome/fontawesome-free": ">=5.15.0 <7.0.0",
-    "@hotwired/turbo-rails": "^7.1.0",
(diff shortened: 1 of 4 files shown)


mshibuya:
Thank you!

PR #3738, opened 2026-04-11
Fix `_discard` leaking into SQL for multi-select enum filters

tonekk:
## Summary

When an enum filter is used in multi-select mode, `_discard` (the `"..."` placeholder) gets submitted as a literal value on subsequent page loads, producing SQL like:

```sql
WHERE (users.role IN ('_discard','admin','editor'))
```

This PR fixes both the client-side and server-side causes.

## Steps to reproduce

1. Have a model with an enum field (e.g. `role` on `User`)
2. Navigate to the index page and open the filter for that enum field
3. Select multiple values (e.g. `admin` and `editor`) and submit the filter
4. On the filtered results page, **do not touch the filter select** — just click "Filter" (submit) again
5. Observe the generated SQL: `_discard` appears as a value in the `IN (...)` clause

## Root cause

**Client-side** (`src/rails_admin/filter-box.js`): When restoring a multi-select filter on page reload, `_discard` remains implicitly selected because `.prop("multiple", true)` preserves the default selection, and the code only hides the option without deselecting it.

**Server-side** (`lib/rails_admin/abstract_model.rb`): The `_discard` guard only checks scalar equality (`v == '_discard'`), which is always false when `@value` is an array like `['_discard', 'admin', 'editor']`.

## Fix

- **Client-side**: Add `.prop("selected", false)` when hiding special options in multi-select mode
- **Server-side**: Strip `_discard` from array values as defense-in-depth; return nil if the array becomes empty after stripping

### DIFF
--- a/lib/rails_admin/abstract_model.rb
+++ b/lib/rails_admin/abstract_model.rb
@@ -140,6 +140,10 @@ module RailsAdmin
       end
 
       def to_statement
+        if @value.is_a?(Array) && @value.include?('_discard')
+          @value = @value.reject { |v| v == '_discard' }
+          return if @value.empty?
+        end
         return if [@operator, @value].any? { |v| v == '_discard' }
 
         unary_operators[@operator] || unary_operators[@value] ||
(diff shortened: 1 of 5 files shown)


mshibuya:
Thanks!

PR #3739, opened 2026-04-15
Declare JavaScript package is a module

jdufresne:
The package entry point uses the import statement, so it is using module syntax.


mshibuya:
👍

PR #3741, opened 2026-08-02
Improve intermittent autocomplete test failures due to stale element

jdufresne:
The jquery UI select widget tests intermittently fail in CI. This is an attempt to fix them by using a more complete and accurate selector. I believe the previous version could get into scenarios where the less specific selector matched a stale element on the page prior to the asynchronous fetch completing.


mshibuya:
Yeah I was aware of the flakiness. Thank you!

PR #3742, opened 2026-08-02
Remove @babel/runtime as an unnecessary direct dependency

jdufresne:
The project doesn't import `@babel/runtime` or use it directly, so remove it as a dependency to reduce downstream version conflicts and node_modules bloat. For example, projects wishing to upgrade to Babel 8.


mshibuya:
👍

PR #3749, opened 2026-08-29
Fully specify all ESM imports with an extension

jdufresne:
In some build environments, the extension is mandatory. Without it, I see errors like:

```
ERROR in rails_admin/src/rails_admin/base.js 28:0-19
Module not found: Error: Can't resolve './widgets' in 'rails_admin/src/rails_admin'
Did you mean 'widgets.js'?
BREAKING CHANGE: The request './widgets' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
resolve './widgets' in 'rails_admin/src/rails_admin'
  using description file: rails_admin/package.json (relative path: ./src/rails_admin)
    Field 'browser' doesn't contain a valid alias configuration
    using description file: rails_admin/package.json (relative path: ./src/rails_admin/widgets)
      Field 'browser' doesn't contain a valid alias configuration
      rails_admin/src/rails_admin/widgets doesn't exist
```

### DIFF
--- a/src/rails_admin/base.js
+++ b/src/rails_admin/base.js
@@ -1,7 +1,7 @@
 import Rails from "@rails/ujs";
 import "@hotwired/turbo-rails";
-import "./jquery";
-import "./vendor/jquery_nested_form";
+import "./jquery.js";
+import "./vendor/jquery_nested_form.js";
 import "bootstrap";
 
 // These jQuery-UI indirect dependencies need to be preloaded to be used within Import maps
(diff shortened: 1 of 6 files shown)


mshibuya:
👍

PR #3750, opened 2026-08-29
Remove duplicate global assignment of window.jQuery

jdufresne:
The window.jQuery global variable is already assigned in
src/rails_admin/jquery.js, so the one in the head template is an unnecessary duplicate.


mshibuya:
~Allow me to close this, as this will be fixed by #3751 anyway 🙏~
Merging this in first 👍

PR #3751, opened 2026-08-30
Ship a prebuilt asset bundle; collapse config.asset_source (4.0)

mshibuya:
## Approach

RailsAdmin carried 5 asset-delivery modes (sprockets / webpacker / webpack / importmap / vite), each with its own installer path, `<head>` branch, and — for Sprockets — vendored copies of Bootstrap, jQuery UI and flatpickr. Webpacker is EOL and the modes kept drifting.

This builds `rails_admin.{js,css}` from `src/` (esbuild + Dart Sass) and **commits them to the gem**. `config.asset_source` becomes:

- `:propshaft` / `:sprockets` — serve the shipped bundle, no build step, no Node
- `:external` — you build it (esbuild / Webpack / Vite / …)
- a callable, or `nil` to auto-detect

Deps that used to be per-mode chores are bundled: `@rails/activestorage` + `@rails/actiontext` glue, and every flatpickr locale. Trix stays CDN-loaded unless you opt into `:external`. Customization with no build step is now the `_head_custom` partial plus `--ra-*` CSS custom properties.

Net **−52k lines** — mostly the vendored Bootstrap/jQuery-UI forks, plus the Sprockets manifests, `ESModuleProcessor`, `@rails/ujs`, `sassc-rails` handling and the Font Awesome `.ttf`.

## Breaking changes (4.0)

- `config.asset_source = :webpacker` / `:vite` now raise; `:webpack` → `:external`, `:importmap` → detected pipeline (both with a deprecation warning)
- `app/assets/…/rails_admin/custom/*` overrides and `rails_admin/application.{js,css}` are gone → use `_head_custom` / `--ra-*` / `:external`
- Sprockets apps move off a vendored Bootstrap 5.1 fork onto 5.3 (minor visual shifts)
- ActionText without a build: Trix loads from a CDN (attachment uploads unaffected)
- `@rails/ujs` no longer bundled — old `link_to …, method: :delete` links rendered inside the RailsAdmin layout need `data: { turbo_method: :delete }` or the host's own rails-ujs
- Font Awesome ships woff2-only

Full details: `docs/upgrading-to-4.md`.

## Reviewing

Every commit is a self-contained change (build toolchain → asset_source rework → one deletion per commit → FA / theming / ujs → docs & CI). Reviewing commit-by-commit is the intended path.


mshibuya:
@jdufresne 
Please forgive me for opening a big AI-assisted PR 🙇‍♂️ 
I want your feedback on the approach, as you know much more about frontend than me!

jdufresne:
This is great! I really like this idea and approach and would like to see this through, however, in the current state, the review is a bit daunting to take on all at once. What do you think about breaking this down into reviewable PRs that can land independently to facilitate that work? For example, I recommend re-opening  #3750 and to create other tickets like that to chip away at problem in easier to read chunks. Are you open to that idea? Even the individual commits here might be good candidates for that approach.

mshibuya:
Thank you for the opinion! I can merge #3750 first, once I fix the master build.
However, I don't think splitting the commits into PRs won't help. If those commits aren't reviewable one by one, how will the resulting PRs be?

Can't we be aligned on the approach (pre-building assets, choice of asset sources to keep, ...) first? Then I want to get the implementation working in master, and make adjustments as necessary. That should be easier for both of us I think 👍

mshibuya:
Please forgive me to go ahead with merging, as the changes need to be field-tested 🙇‍♂️ 
If you have any feedbacks I'm happy to address. Please feel free to leave comments anytime!

jdufresne:
> Please forgive me to go ahead with merging, as the changes need to be field-tested 🙇‍♂️ If you have any feedbacks I'm happy to address. Please feel free to leave comments anytime!

Sounds good. Yes please go ahead. I'm afraid I won't have the time to commit to this that it deserves.

